PR #20, opened 2025-01-13
chg: add sentry to Q cluster and mark shas with pragma to ignore

larry6point6:
This should ensure worker sends logs to Sentry


larry6point6:
Change made in more recent PR

PR #30, opened 2025-01-21
fix: add pagination to API

nboyse:
## Context

We have way too much data on dev and prod for it to handle just a simple curl command.
This allows pagination on the API in which lets us actually connect to our endpoint without a 504

## Guidance to review
Check you're happy with my approach

## Things to check


a-i-flo (approved):
Pagination works and data is correct. Still left to add the model.

PR #36, opened 2025-01-22
feat: complete serializer for user details

nboyse:
## Context

We would like a table in data workspace that reflects all user details we have so far without exposing PII

Adds all the missing fields for the serializer

## Guidance to review
Check I didnt miss anything

## Things to check


a-i-flo (approved):
Comments (not sure if makes a difference)
- id and email is in the first fields but not second
- chats is in the second list

PR #40, opened 2025-01-24
chg:wrap google analytics_iframe_src as a string

larry6point6:


nboyse (approved):

django_app/redbox_app/templates/base.html
```diff
@@ -40,7 +40,7 @@
 
   {% if environment | lower == "prod" and waffle_flag(request, 'uktrade') %}
   <!-- Google Tag Manager (noscript) -->
-  <noscript><iframe src={{ google_analytics_iframe_src }}
+  <noscript><iframe src="{{ google_analytics_iframe_src }}""
```
nboyse, line 43: Ah! Extra "

nboyse (commented):
ah

django_app/redbox_app/templates/base.html
larry6point6: Ah good catch updated :)

PR #43, opened 2025-01-24
update CSP with new SHA

larry6point6:


nboyse (approved):
All good - shall we make a ticket in the backlog to review the sha's at a later date? Worried about security etc.

PR #48, opened 2025-01-28
chg: add google analytics iframe variable to jinja2.py and settings.py

larry6point6:
One of the variables for google analytics was not being propagated through, added it so should now pull through


larry6point6:
No longer needed been covered in another PR

PR #60, opened 2025-02-03
chg: add google reccomeded connect-src CSP

larry6point6:
Updating CSP based on errors in prod


django_app/redbox_app/settings.py
```diff
@@ -223,6 +223,9 @@
     f"{WEBSOCKET_SCHEME}://{ENVIRONMENT.hosts[0]}/ws/chat/",
     "eu.i.posthog.com",
     "eu-assets.i.posthog.com",
+    "https://*.google-analytics.com",
+    "https://*.analytics.google.com",
+    "https://*.googletagmanager.com",
     "https://www.google-analytics.com/",
     "https://region1.google-analytics.com/",
     "https://www.googletagmanager.com/",
```
larry6point6, line 231: Unsure if we still need lines 229-231 here because the above should enable all variants of GTM to be reachable

nboyse (approved):

PR #78, opened 2025-02-11
test: fix API tests

michalc:
## Context

There are some broken tests from some legacy work. This fixes just some of them: the ones in test_api_views.py

There were 3 issues causing the tests to fail that this change fixes:

- An incorrect assumption on the form of an error message in the case of a 403
- Using a now-removed date_joined field
- Passing an email address into model creation, where is also determined the email address based on username.

## Guidance to review

Looking at the code and making sure more tests don't fail should be enough.

## Relevant links

N/A

## Things to check


michalc:
Closing to rename the branch to get tests to run in GitHub actions

PR #80, opened 2025-02-11
Fix summarisation and exponential backoff for summarisation

nora-errouhly:
- fixed the summarisation prompt to prevent LLM getting confused as it summarise entire document and not just the last section of the document
- fixed ThrottlingException limit issues for summarisation using exponential backoff (implemented in the 3 summarisation nodes).


nboyse (approved):
Nice.

PR #94, opened 2025-02-17
fix: implement exponential backoff to resolve EventStreamError

nboyse:
## Context

We are getting numerous flags from sentry about:

`EventStreamError: An error occurred (serviceUnavailableException) when calling the InvokeModelWithResponseStream `

We dont want this anymore. Its probably because too many requests are being sent to bedrock, so exponential backoff is a healthy way to try mitigate this error.

Implement exponential backoff. Maybe further in future we should look at other errors to catch (ThrottlingException etc.)

## Guidance to review

Check it is a suitable implementation

## Things to check

### DIFF
--- a/redbox-core/redbox/graph/nodes/processes.py
+++ b/redbox-core/redbox/graph/nodes/processes.py
@@ -2,10 +2,14 @@ import json
 import logging
 import re
 import textwrap
+import time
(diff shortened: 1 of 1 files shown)


larry6point6 (approved):
LGTM!

PR #98, opened 2025-02-18
fix: Remove tiktoken references

nboyse:
## Context

Tiktoken is making direct calls to openAI which is causing a RateLimitError as we don't use openai anymore we shouldn't even be referencing it (security wise not great either)

Its only for metadata so doesn't impact anything properly but turning it off will reduce the amount of herrings and ensure our independence.

Replace tiktoken references with a custom tokeniser simply used to count token usage across codebase and minorly used for wikipedia/govuk search functions

## Guidance to review

Please run it locally, check it hasn't broken any of the routes for you, also check you are happy with the approach or if you have any suggestions on what you'd rather use.

## Things to check


larry6point6:
As discussed think the implementation is all good but not sure if right, ultimately a question of how important are tokens being accurate per model basis and the impact either way, think it's for the team to decide/guide

PR #100, opened 2025-02-19
fix: prevent 504 for user table

nboyse:
## Context

Following on from DIAPPs work ([here](https://github.com/uktrade/redbox/pull/76)) we would like to ensure the errors stop for our other two tables too. DIAPP has now covered the information we need regarding Chat/ChatMessages so I'm severing the connection between chats in the UserSerializer

Remove ChatSerializer as its covered by ChatMessageSerializer
Add extra fields to UserSerializer
Remove chats from being called unnecessarily in the UserSerializer which is causing the 504 timeouts
Update relevant tests


a-i-flo (commented):
Needs a column with user_id to be able to link with the chat_message_details.

a-i-flo (approved):

PR #103, opened 2025-02-20
feat: redirect user trying to delete failed document upload

nboyse:
## Context

Currently, documents which have errored cannot be deleted. Or at least, they are "technically" deleted but because they are already removed from opensearch users are seeing a random error, when they are automatically deleted.

Just redirect the user so they don't see a random false flag regarding inactive file uploads. Its not important to them as if a doc fails to upload, it will automatically delete.

## Relevant links
https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-500

## Things to check


larry6point6 (approved):
LGTM!

PR #118, opened 2025-02-26
Bugfix/search irrelevant query

nora-errouhly:
removing BM25 and keeping semantic search only


nora-errouhly:
testing:
- run redbox via docker compose or locally
- through the UI , upload the document power_chapter.pdf
- enter the prompt "what is the door about" and check that LLM does not hallucinate

larry6point6 (approved):
LGTM! Tested locally looks good and works as described

PR #132, opened 2025-03-05
Chore/update env variables

larry6point6:
Error happening occasionally in sentry, added env variable to solve https://sentry.ci.uktrade.digital/organizations/dit/issues/154382/?environment=dev&project=272&query=is%3Aunresolved


django_app/redbox_app/settings.py
```diff
@@ -463,4 +463,4 @@ def filter_transactions(event, _hint):
 CLAM_AV_USERNAME = env.str("CLAM_AV_USERNAME", " ")
 CLAM_AV_PASSWORD = env.str("CLAM_AV_PASSWORD", " ")
 CLAM_AV_DOMAIN = env.str("CLAM_AV_DOMAIN", " ")
-USE_CLAM_AV = env.str("USE_CLAM_AV")
```
nboyse, line 466: if we are removing this bool do we not also need to remove from .env.example? and create relevant deploy PR?

larry6point6: Ah I had it in twice as a bool higher up and string there, so removed the string version as it wasn't needed

PR #146, opened 2025-03-11
feature: microphone implementation

nboyse:
## Context

We want the user to have the ability to record themselves in order to interact with the LLM

## What is this code aiming to achieve?
Implements the functionality required in order for the user to use a microphone if so wanted

- [ ] No
Probably a good idea to run it locally and check all is well and it can understand you

## Relevant links
https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558/backlog?selectedIssue=REDBOX-670


larry6point6 (approved):
LGTM

PR #155, opened 2025-03-19
feat: implement auto reload with uvicorn

nboyse:

- [ ] No


nboyse:
We have found an otherwise suitable alternative to auto reload

PR #158, opened 2025-03-20
set OpenSearch min score RAG to 6

saisakul:

- [ ] No


nora-errouhly (approved):
approved

PR #178, opened 2025-04-03
Add citation to search

saisakul:
## Context

Inline citations help user understand where from the source the LLM derives answer from, and also create trusts.

## What is this code aiming to achieve?

Show inline citations when using self_route and @search.

- [ x] No

Test in docker to see the response return with citations.

### DIFF
--- a/redbox-core/redbox/chains/runnables.py
+++ b/redbox-core/redbox/chains/runnables.py
@@ -8,14 +8,7 @@ from langchain_core.messages import AIMessage, AIMessageChunk, BaseMessage
 from langchain_core.output_parsers import StrOutputParser
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
 from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
-from langchain_core.runnables import (
-    Runnable,
-    RunnableBranch,
-    RunnableGenerator,
-    RunnableLambda,
-    RunnablePassthrough,
-    chain,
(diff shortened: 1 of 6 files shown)


redbox-core/redbox/models/prompts.py
ugo-daintta, line 1: Hello, I find that if not explicitly told to, LLMs will default to US spelling. Can we preserve the "Use UK English spelling in response." section in the prompts?

saisakul: can we make this a separate ticket?

redbox-core/redbox/chains/runnables.py
ugo-daintta, line 1: Can I check why you made `parser` nonlocal in the `_self_route_output_parser` method (line 159)? It looks like it wasn't updated in this method?

redbox-core/redbox/models/prompts.py
ugo-daintta: Can do. But was it leading to errors hence you deleting the section about UK spelling from the current version of the prompt?

ugo-daintta: Also it's worth saying that I have not spotted any US spelling in testing so far. So it might not be necessary 😄

redbox-core/redbox/chains/runnables.py
saisakul: updated.

ugo-daintta (approved):
Many thanks. LGTM

PR #188, opened 2025-04-10
Feature/citation links

ugo-daintta:
## Context
When a question is asked via the @search route (directly or through self-direct) with at least one document selected. The response includes a footnotes section with citations.

**Current Behaviour** : When clicked, the citations link to the footnote section on the same page.

**Proposed Change**: Instead, this would like to the sources page with emphasis on the text from the link that has been selected.

## What changed
-  Updated the `unique_citation_uris` method in `django_app/redbox_app/redbox_core/models.py:856` to return the `citation.id` which is necessary for emphasis on the summary screen
- Updated `django_app/redbox_app/redbox_core/views/chat_views.py:58` to return a link to the sources summary (instead of the footnotes section using the now provided `citation.id`
-  Added `citation.id` as an identifier in `django_app/redbox_app/templates/citation_fragment.html` to allow for selecting the relevant section of the sources page.
- Updated `django_app/redbox_app/templates/macros/chat-macros.html:59` to avoid an error due to the update to the `unique_citation_uris` method.

- [ ] No
Run locally and confirm that this does in fact link to the correct section.

## Relevant links
[JIRA Ticket - Changing the structure of citations](https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?assignee=712020%3Ac1df826f-3b55-414a-9586-d302d6fb4f88&selectedIssue=REDBOX-725)

### DIFF
--- a/django_app/redbox_app/redbox_core/models.py
+++ b/django_app/redbox_app/redbox_core/models.py
@@ -853,8 +853,8 @@ class ChatMessage(UUIDPrimaryKeyBase, TimeStampedModel):
             body=elastic_log_msg,
         )
 
-    def unique_citation_uris(self) -> list[tuple[str, str]]:
-        """a unique set of names and hrefs for all citations"""
+    def unique_citation_uris(self) -> list[tuple[str, str, str, str]]:
+        """a unique set of names, hrefs, ids and relevant texts for all citations"""
 
         def get_display(citation):
             if not citation.file:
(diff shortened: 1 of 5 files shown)


django_app/redbox_app/redbox_core/models.py
```diff
@@ -546,6 +546,9 @@ class Status(models.TextChoices):
     def __str__(self) -> str:  # pragma: no cover
         return self.file_name
 
+    def __lt__(self, other):
+        return self.id < other.id
```
nboyse: Why has this been moved

nboyse: have moved it back as it was part of the formatting errors

nboyse (approved):
LGTM and formatting fixed

django_app/redbox_app/redbox_core/models.py
ugo-daintta: Ruff commanded and I followed 😄

PR #192, opened 2025-04-15
ENV var for LLM max tokens

saisakul:
## Context

LLM_max_tokens are stored in various places, causing confusion. 

## What

Use ENV to store llm max tokens setting.

- [ ] No

### DIFF
--- a/.copilot/image_build_run.sh
+++ b/.copilot/image_build_run.sh
@@ -3,4 +3,5 @@
 set -e
 
 # Add commands below to run inside the container after all the other buildpacks have been applied
-python manage.py collectstatic --noinput --clear
+cd django_app
+poetry install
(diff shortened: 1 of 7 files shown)


redbox-core/redbox/models/settings.py
```diff
@@ -85,8 +85,6 @@ class Settings(BaseSettings):
     embedding_backend: str = "amazon.titan-embed-text-v2:0"
     embedding_backend_vector_size: int = 1024
 
-    llm_max_tokens: int = 1024
```
nboyse, line 88: should this be like above too?
`llm_max_tokens: int = env.int("LLM_MAX_TOKENS", default=1024)`

saisakul: I don't see this gets used anywhere so removed it.

nboyse (approved):

PR #197, opened 2025-04-16
Feature/improve evaluator

nora-errouhly:
- changing graph (newroute state management)
- changing evaluator and planner prompts
- adding env variable to switch on/off citations in new route. value is set to False in env file (switching off citations in newroute)


redbox-core/redbox/chains/runnables.py
```diff
@@ -72,16 +72,18 @@ def _chat_prompt_from_messages(state: RedboxState) -> Runnable:
             | _additional_variables
         )
 
-        return ChatPromptTemplate(
+        chatprompt = ChatPromptTemplate(
             messages=(
                 [("system", system_prompt_message)]
                 + [(msg["role"], msg["text"]) for msg in truncated_history]
                 + [MessagesPlaceholder("messages")]
-                + [task_question_prompt + "\n\n{format_instructions}"]
```
saisakul, line 80: this affects other routes, please modify prompts for other routes to include format_instructions.

saisakul:
I don't think we should create a separate route for response without citations. It should use the same route but can handle different settings.

redbox-core/redbox/chains/runnables.py
nora-errouhly: can you specify which route/ prompts it may affects. I can then append the prompt with format instructions. any prompt modification should be in prompts.py, not in runnables.py

nora-errouhly: I can see format_instructions are in many system prompts. if they are defined in system prompts, they should not be required in question prompts

nora-errouhly: I added format instruction to one of the prompts. all other prompts are fine

nora-errouhly:
> I don't think we should create a separate route for response without citations. It should use the same route but can handle different settings.

I did not add another route. it is still within same @newroute

ugo-daintta (approved):

PR #206, opened 2025-04-22
remove evaluation part from evaluator

saisakul:
## Context

The evaluator is currently doing:
1) answer user question 
2) provide citation
3) evaluate whether answer satisfies user question/intention

which is too complex for a single agent. 

## What

Change the evaluator prompt so that it only answers questions with citations

- [x] No

1. Run Redbox app in docker
2. Test using @newroute your question
3. Redbox should return response without error.

### DIFF
--- a/redbox-core/redbox/models/prompts.py
+++ b/redbox-core/redbox/models/prompts.py
@@ -40,19 +40,6 @@ RETRIEVAL_SYSTEM_PROMPT = """
 
    """
 
-EVALUATOR_PROMPT = """
-1. Comprehensive Analysis:
-   - Evaluate question against full message history
-   - Assess contextual completeness
-   - Determine answer feasibility
-
-2. Quality Criteria:
-   - Technical accuracy
-   - Depth of explanation
-   - Contextual relevance
-   - Clarity of communication
-   """
-
 NEW_ROUTE_RETRIEVAL_SYSTEM_PROMPT = """Answer user question usinng responses received from other AI agents. Use citations to back up your answer. Return in the format <Format>{format_instructions}</Format>
 """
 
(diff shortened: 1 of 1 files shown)


redbox-core/redbox/models/prompts.py
```diff
@@ -40,8 +40,7 @@
 
    """
 
-NEW_ROUTE_RETRIEVAL_SYSTEM_PROMPT = """Expert Answer Evaluation Protocol:
-
+EVALUATOR_PROMPT = """
```
nora-errouhly, line 43: is evaluator prompt used elsewhere in the code?

redbox-core/redbox/models/prompts.py
```diff
@@ -263,7 +241,7 @@
 
 RETRIEVAL_QUESTION_PROMPT = "<User question>{question}</User question>"
 
-AGENTIC_RETRIEVAL_QUESTION_PROMPT = "{question}"
+AGENTIC_RETRIEVAL_QUESTION_PROMPT = "<User question>{question}</User question>"
```
nora-errouhly, line 244: do we need these tags? usually tags helps structure long prompt but this prompt is short anyway

redbox-core/redbox/models/prompts.py
saisakul: good spot. Will remove.

redbox-core/redbox/models/prompts.py
saisakul: according to https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/use-xml-tags, using tag helps Claude parse your prompts more accurately, leading to higher-quality outputs.

richiebird (approved):

PR #213, opened 2025-04-25
feat: data hub api functionality initial step

nboyse:
## Context

We would like to be able to access data hub's various datasets from Redbox. Work has been done via SRE to ensure VPCs can be communicated between each other in dev, so now we have to do the first step, which is the function that allows claude to process and understand data hub's dataset structure.

## What

Prompt engineering, filter matching and then a main function which communicates between data hub's API and redbox. It is important to note this should only be used on Dev at this stage, so we have an appropriate waffle flag around each function to ensure we do not use it in prod (it wouldn't work anyhow as VPC peering has not been turned on here *subject to approval from IRAP this will change*)

- [ ] No
- [X] Maybe

I have carried out a fair amount of testing in the initial PR [here](https://github.com/uktrade/redbox/pull/200/files#diff-f88980af4e12a3aed23231673bc1bebc63956208fd2781a484a8fc4ddc3d0987R155) and I am happy with the results, i.e., claude can understand data hub's various datasets. These tests can be used locally, after fetching the secrets from AWS, if you would also like to verify the expected output is met. But the main testing will be carried out and written once the agent is created. This is just a first step.

## Relevant links
https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-749


nora-errouhly (approved):
approved. we'll do more testing once the new agent is created

PR #217, opened 2025-04-28
Udating prompt to ensure correct format with citations is returned

saisakul:
## Context

Updating the prompt in newroute by giving examples of expected output.

- [ ] No

Run your Redbox in local app, select a document with has author info or use Power Chapter and test if 
test 1: response contains citations
prompt = @newroute who wrote the document?
expected outcome: response returned with at least 1 citation

test 2: response contains no citations
prompt = @newroute what has been mentioned about Saisakul?
expected outcome: response returned with no citation

### DIFF
--- a/redbox-core/redbox/models/prompts.py
+++ b/redbox-core/redbox/models/prompts.py
@@ -41,9 +41,11 @@ RETRIEVAL_SYSTEM_PROMPT = """
    """
 
 
-NEW_ROUTE_RETRIEVAL_SYSTEM_PROMPT = """Answer user question using the provided context. Use citations to back up your answer.  {format_instructions}. 
-For example, your JSON output shoud be in the format: "answer": your_answer, "citations": citations
-DO NOT start by saying: 'Here is the JSON response', just return JSON."""
+NEW_ROUTE_RETRIEVAL_SYSTEM_PROMPT = """Answer user question using the provided context. Use citations to back up your answer when available. Always return your response in the following JSON format: {{"answer": your_answer, "citations": [list_of_citations]}}.
+
+If no citations are available or needed, return an empty array for citations like this: {{"answer": your_answer, "citations": []}}.
+
+{format_instructions}"""
 
 AGENTIC_RETRIEVAL_SYSTEM_PROMPT = (
     "You are an advanced problem-solving assistant. Your primary goal is to carefully "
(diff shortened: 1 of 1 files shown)


redbox-core/redbox/models/prompts.py
```diff
@@ -41,9 +41,11 @@
    """
 
 
-NEW_ROUTE_RETRIEVAL_SYSTEM_PROMPT = """Answer user question using the provided context. Use citations to back up your answer.  {format_instructions}. 
-For example, your JSON output shoud be in the format: "answer": your_answer, "citations": citations
-DO NOT start by saying: 'Here is the JSON response', just return JSON."""
+NEW_ROUTE_RETRIEVAL_SYSTEM_PROMPT = """Answer user question using the provided context. Use citations to back up your answer when available. Always return your response in the following JSON format: {{"answer": your_answer, "citations": [list_of_citations]}}.
+
+If no citations are available or needed, return an empty array for citations like this: {{"answer": your_answer, "citations": []}}.
```
nora-errouhly, line 46: this seems good. on the last sentence, you add instructions to put citation as empty list. however, this is already specified in transform.py so I don't think the last sentence is needed. what do you think?
here is the function in transform.py
@RunnableLambda
def get_all_metadata(obj: dict):
    text_and_tools = obj["text_and_tools"]

    if parsed_response := text_and_tools.get("parsed_response"):
        text = getattr(parsed_response, "answer", parsed_response)
        citations = getattr(parsed_response, "citations", [])
    else:
        text = text_and_tools["raw_response"].content
        citations = []

    out = {
        "messages": [AIMessage(content=text, tool_calls=text_and_tools["raw_response"].tool_calls)],
        "metadata": to_request_metadata(obj),
        "citations": citations,
        "final_chain": obj["final_chain"],
    }
    return out

redbox-core/redbox/models/prompts.py
```diff
@@ -235,7 +213,9 @@
 AGENTIC_RETRIEVAL_QUESTION_PROMPT = "<User question>{question}</User question>"
 
 NEW_ROUTE_RETRIEVAL_QUESTION_PROMPT = (
-    "<User question> {question} </User question> \n\n <Context>: \n\n {agents_results} \n\n </Context> \n\n Answer:"
+    "<User question> {question} </User question> \n\n <Context>: \n\n {agents_results} \n\n </Context> \n\n."
+    "Put it in <sonnet> tags.\n"
```
nora-errouhly, line 217: I don't understand the <sonnet> tag. what is it?
I think "Assistant" is only required when used in conjunction with "Human".
do we not need "Answer" at the end?

saisakul: I got inspiration from here https://docs.google.com/spreadsheets/d/19jzLgRruG9kjUQNKtCg1ZjdD6l6weA6qRXG5zLIAhC8/edit?gid=2007125861#gid=2007125861

By putting `Assistant: <sonnet>` it ensures that response starts with <sonnet>. I have tested removing it and Claude will return preamble.

redbox-core/redbox/models/prompts.py
saisakul: This is to ensure that the LLM knows what to return when there is no citation available, preventing LLM returning just answer i.e. just text. For example:
```
LLM return: 'There is no information on this topic'
```

VS 

```
LLM return: {'answer': 'There is no information on this topic', 'citations': []}
```

redbox-core/redbox/models/prompts.py
nora-errouhly: I see, thanks for sharing the link. did you see any benefit by adding the sonnet tag?

saisakul: yes, as before even when I specified to the LLM to not return other things beside JSON, the LLM still does (sometimes).

PR #220, opened 2025-04-29
add max_tokens to agents

saisakul:
## Context

There is no token limit for agents hence in some cases agents could return a very long response (large #tokens) causing error: `EventStreamError('An error occurred (validationException) when calling the InvokeModelWithResponseStream operation: Input is too long for requested model.`

## What

- add max tokens limit per agents
- truncate the response so that it does not exceed token limit
- update gov.uk tool so that it only fetch first 1000 tokens of contents

- [ ] No

https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-757
1. Spin up local Redbox
2. Using this prompt: @newroute summarise from the perspective of a policy analyst working in the department for business and trade and find the original report on gov uk
Document: ar24_e (1).pdf 
3. Redbox produces response without error.


richiebird (approved):

richiebird:
Correction: Testing instructions step 3 should read 
"Redbox produces response **without** error."

PR #221, opened 2025-05-01
Feature/citation streaming

ugo-daintta:
## Context

Currently when streaming, citations do not appear reliably with one or no citations appearing when the question is initially asked and only visible when refreshed.

The goal of this PR is to ensure that citations are included in the streamed response reliably and pointing to the `citations` tab. The further step requiring highlighting will be included in a subsequent repo.

# What's been changed
1. **Changed uniqueness criteria in `sources-list.js` to be based on citation text instead of file name.** Previously, citations would only stream if there was a single source from a unique file. This change allows for multiple citations from the same file to be streamed

2. **Added in the `escapeHtmlAttibute` method to `sources-list.js`**. This method escapes quotation marks and other symbols that were causing a html query method to break. Previously, if the citation text had quotation marks this would cause a conflict with the html tag when a downstream get command is called in `chat-message.js`. Now we convert these to special symbols which do not cause the conflict.

3. **Added `chatId` as an argument to the addFootnotes method and modified the link replacement in the innerHTML**. This allowed for the link to point to the citations tab and not the sources below. Currently the footnote id is put in as a placeholder but it needs to be replaced with the citation ID once it's gleaned from the DOM.

- [ ] No

- Ask the models question using the search keyword that require citations.
- Confirm that the complete citations are generated with links that go to the `citations` tab
- When the response tab is refreshed, new citations do not appear


nora-errouhly:
I tried 3 different questions but I am getting an empty response. not sure if this is related to this work or is this a separate bug. Can you please check? also please can you send me a working example that you have tested so I can test it as well. 
Screenshot 2025-05-01 at 11 54 43
Screenshot 2025-05-01 at 11 56 54

ugo-daintta:
Interesting. Maybe try the below?
image

ugo-daintta:
Link to the pdf https://assets.publishing.service.gov.uk/media/6645c709bd01f5ed32793cbc/Green_Book_2022__updated_links_.pdf

nora-errouhly:
I tested with search and newroute. the following test with newroute has a missing citations (it is in the source but not in the text). can you check why is this happening? All citations appear without the need to refresh but the number of sources do not match the citations references within the text
Screenshot 2025-05-01 at 12 46 14

nora-errouhly:
I tested and refreshed the page. after refreshing, the order of citations change randomly which might affect the accuracy of citations (have seen a wikipedia source being changed to document source but it is not shown in these pictures). attached picture show difference after refreshing pages.
Screenshot 2025-05-01 at 13 17 57
3-22b1-4d41-97cc-d5c9cce48608" />
 currently @search route give a lot of empty responses so have been testing more with newroute.

ugo-daintta:
Thanks Nora. The order changes but the links stay the same No? i.e. Text from source 1 is marked with a 1 and when refreshed is marked with a 3 and replaced with 3? Unless I'm not understanding?

ugo-daintta:
Latest push will handle citation ordering issue

django_app/frontend/src/js/web-components/chats/chat-message.js
```diff
@@ -247,15 +247,15 @@ export class ChatMessage extends HTMLElement {
...
-        this.#addFootnotes(streamedContent);
+        this.#addFootnotes(streamedContent,response.data.message_id);
         const chatResponseEndEvent = new CustomEvent("chat-response-end", {
           detail: {
             title: response.data.title,
             session_id: response.data.session_id,
           },
         });
         document.dispatchEvent(chatResponseEndEvent);
+        location.reload()
```
nboyse, line 258: has this finished it all off?!

django_app/frontend/src/js/web-components/chats/sources-list.js
```diff
@@ -36,13 +50,16 @@ export class SourcesList extends HTMLElement {
                       source.url
                     }" id="footnote-${this.getAttribute("data-id")}-${
         this.sources.length
-      }" data-text="${source.matchingText}">${source.fileName || source.url}</a>
+      // }" data-text="${source.matchingText}">${source.fileName || source.url}</a>
```
nboyse, line 53: might as well delete this unless we're going to use at later date?

django_app/frontend/src/js/web-components/chats/sources-list.js
```diff
@@ -36,13 +50,16 @@ export class SourcesList extends HTMLElement {
...
+      //           </li>
+      //       `;
+      }" data-text="${this.escapeHtmlAttribute(source.matchingText)}">${source.fileName || source.url}</a>
                 </li>
             `;
     });
     html += `</div></ol>`;
-
     this.innerHTML = html;
+    console.log(html)
```
nboyse, line 62: Ah probably worth removing this too

nboyse (changes requested):

django_app/frontend/src/js/web-components/chats/sources-list.js
ugo-daintta: @nboyse, I've reflected the changes. I've not been able to adequately test the last push, but it's simply a commented out code deletion and a removal of a console log. Should not be a breaking change

nboyse (approved):

PR #222, opened 2025-05-01
feat: allow user to delete all documents

nboyse:
## Context

Allow user to delete all documents

## What

Creates new route and user journey for users to delete every document

- [ ] No

Try delete all your documents


ugo-daintta:
Hello, my local test has not worked as shown below after multiple refreshes.

image

ugo-daintta:
Tested following recent pushes and it worked like a charm :)

ugo-daintta (approved):

PR #225, opened 2025-05-01
Bugfix/citation prompt

saisakul:
## Context

1) Citation-related prompts are found in various system prompt e.g. in retrieval, gadget, etc. We should only have one citation prompt so that it is easy to debug and make changes.
2) By passing citation prompt at the end, helps LLM returns correct format.

## What

- add new citation prompt
- update chat prompt function to include format instruction at the end of the prompt

- [ ] No

- Spin up docker Redbox app
- Test that all routes are working i.e. can produce answer, and no blank response.


ugo-daintta:
Hello, it looks like the search endpoint with a selected document is using external sources without using information from the document provided. I've included some examples below. Is this the intended behaviour?
image
image

ugo-daintta:
Thanks, I have tested the different endpoints (roughly 2 queries per endpoint). Did not experience any blank responses. Happy to approve

ugo-daintta (approved):

PR #233, opened 2025-05-07
Feature/null response log

ugo-daintta:
## Context
When the LLM sends a null response, no error is raised and this is not captured in the logger or Sentry

## What
Checks that the LLM response is not null. If so, it raises an exception which is sent to the logger and the client. The note to the client shows the end user that there was an error

- [ ] No


nora-errouhly:
I tested the changes in dev and I got the LLM response error in sentry as expected.
Screenshot 2025-05-14 at 10 33 18
Screenshot 2025-05-14 at 10 32 52

nora-errouhly (approved):
approved as error appeared in sentry when testing

PR #236, opened 2025-05-09
feat: use unstructured chunks for metadata

nboyse:
## Context

At the moment 1mb == ~8 seconds
Target 50x1mb == ~60 seconds
Reduction factor ~6.66
New target ~1.2 seconds x 1 mb doc

## What

Refactoring the ingest file functionality and the appropriate loaders.

Reusing UnstructuredChunkLoader chunking for metadata extraction. Instead of MetadataLoader redundantly chunking the file, UnstructuredChunkLoader.lazy_load is going to generate chunks once, which are passed to MetadataLoader.extract_metadata.

- [ ] No

Upload your docs, how fast does it load? Do all types that you'd normally expect to work, work?

## Relevant links
https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-785


ugo-daintta:
Uploaded 4 documents on the `use-unstructured-chunks-for-metadata` branch and then uploaded the same documents on a branch without the refactoring. Total document size was roughly 4.8MB. Total upload time was 1 min 23s vs 3 min 40s albeit with upload failures and thus retries in the second case. 

Refactoring appears to have made the upload faster in local implementations.

ugo-daintta (approved):

PR #238, opened 2025-05-12
Feature/evaluator question

nora-errouhly:
fix bug in question passed to evaluator
test newroute using UI

Note: Please note that UI returns sometimes blank responses. this is not due to newroute but it is an issue with the app. If you test using a notebook, newroute consistently returns a  non-empty response


redbox-core/redbox/graph/root.py
```diff
@@ -857,8 +857,7 @@ def build_new_graph(
     )
 
     builder.add_node("Evaluator_Agent", create_evaluator())
-    builder.add_node("clear_tasks", delete_plan_message())
-    builder.add_node("pass_user_prompt_to_LLM_message", build_passthrough_pattern())
+    builder.add_node("combine_question_evaluator", combine_question_evaluator())
```
saisakul, line 860: Instead of having extra node to get task description to tasks_evaluator, why don't we do it in planner? so rather than adding plan to messages, add to tasks_evaluator.

saisakul (changes requested):

redbox-core/redbox/graph/root.py
nora-errouhly: mainly due to different formats of data: we use planner output  (which has a different json output) in sending_task  while we save the string (question) in tasks_evaluator

PR #247, opened 2025-05-16
feat: sticky textbox

nboyse:
## Context

We want the text box to be visible all the time

## What

Makes the text box sticky so it shows up all the time

- [ ] No
Scroll up and down the page, can you still see the text box


ugo-daintta:
ChatBox remains visible during streaming and in different scroll positions once streaming is complete.

image
image

ugo-daintta (approved):

PR #251, opened 2025-05-19
fix: incorrect reference to logo

nboyse:
## Context

We cant see the logo because the path is wrong

## What

Correct the path

- [X] No


ugo-daintta:
Logo appears in footer
image

PR #259, opened 2025-05-21
Add user feedback for replanning

saisakul:
## Context

When the planner creates a plan, we want human to be able to see and approve the plan before they are sent to worker agents.

Planner agent creates a plan. A human can approve or reject the plan. If the plan is approved, the tasks are sent to work agents. If the plan is rejected, an agent re-plan according to human input. The new plan is sent to worker agents.

image

## What

- adding human in the loop for planning
- add new planner parser so it streams tasks created by planner
- modify the graph so that user doesn't need to use keyword to send in feedback

- [ ] No

1. set `ALLOW_PLAN_FEEDBACK` in .env to `True`
2. spin up your local Redbox ap
3. use @newroute to ask question. for example select green book and input @newroute how do we evaluate project?
4. observe that a plan is streamed
5. put in feedback. Try 2 options:
5.1 accept the plan. for example: go ahead with the plan
5.2 ask to modify plan. for example: don't do summarisation.


ugo-daintta:
It seems to be writing 2 summaries in certain cases. Not sure if this is expected behaviour. Will post responses in JIRA

ugo-daintta:
Will approve as discussed :)

PR #272, opened 2025-05-28
Adding back display uniqueness check

ugo-daintta:
## Context
We want sources to appear only once in the sources section of a chat. 

## What

We are adding a check that returns the name in the html if it hasn't already been displayed.

- [ ] No

Confirm that new and previous chats that returned citations with one or more files included only list each file once with citation links being unaffected.

#### Expected Behaviour
**Before**
image
**After**
image

### DIFF
--- a/django_app/redbox_app/templates/macros/chat-macros.html
+++ b/django_app/redbox_app/templates/macros/chat-macros.html
@@ -42,10 +42,14 @@
   <h3 class="iai-chat-bubble__sources-heading govuk-heading-s govuk-!-margin-bottom-1">Sources</h3>
   <div class="iai-display-flex-from-desktop">
     <ol class="rb-footnote-list govuk-!-margin-bottom-0">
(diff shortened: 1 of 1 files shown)


django_app/redbox_app/templates/macros/chat-macros.html
```diff
@@ -54,7 +58,7 @@ <h3 class="iai-chat-bubble__sources-heading govuk-heading-s govuk-!-margin-botto
 </div>
 {% set seen_files = [] %}
 {% for file in message.selected_files.all() %}
-  {% if file not in seen_files %}
+  {% if file.id not in seen_files %}
```
nboyse, line 61: but the file.id differs dependent on the reference being appending at the end, this should be done on a different key that doesnt change

nboyse (changes requested):

nboyse (approved):

PR #278, opened 2025-06-05
Feature/large doc chunking

ugo-daintta:
## Context

When a user uploads a document with many pages even when the document is small (~3mb).  Redbox fails to successfully process the document. The primary source of this issue is with unstructured which times out for large multi-page pdfs
This PR adjusts the upload method for PDFs with >150 pages to fix this issue.

## What

When a document is selected for upload, the number of pages is checked.
1. If this number exceeds 150 pages:
   - The document is split into groups of 75 pages
   - Separate calls to unstructured are then made for these 75-page groups
   - The elements are then aggregated and passed to OpenSearch

2. If not, the document is uploaded via the extant process.
3.  The `bulk_size` argument has also been passed to OpenSearch to raise the value from the default 500 to 1000.

- [ ] No
Try to upload a large document (e.g. 300-999 pages long) and confirm that it is processed successfully


redbox-core/redbox/loader/loaders.py
```diff
@@ -63,30 +88,61 @@ def _get_chunks(self, file_name: str, file_bytes: BytesIO) -> list[dict]:
...
+                    msg = f"Chunk {idx+1} failed: {response.text}"
+                    print(response)
+                    raise ValueError(msg)
+                elements.extend(response.json())
+        else:
+            files = {
+                "files": (file_name, file_bytes),
+            }
+
+            response = requests.post(
```
saisakul, line 125: can we reduce the duplicate codes here (125) and line 102?

ugo-daintta: Good point! Have refactored

PR #314, opened 2025-06-17
Bugfix/validation error

saisakul:
## Context

Sometimes LLM returns data that does not conform with our `Source` class causing `ValidationError`.

https://uktrade.atlassian.net/browse/REDBOX-894?atlOrigin=eyJpIjoiNzBiNmYxOGVmMjZkNDFiZTlhNmIwMjZjZDJkNmUxYzUiLCJwIjoiaiJ9

## What

- Adding validation check to update the value to conform with the `Source` class
- Add test case to test invalid LLM citation output

- [ ] No

### DIFF
--- a/redbox-core/redbox/models/chain.py
+++ b/redbox-core/redbox/models/chain.py
@@ -11,7 +11,7 @@ from langchain_core.documents import Document
 from langchain_core.messages import AnyMessage
 from langgraph.graph.message import add_messages
 from langgraph.managed.is_last_step import RemainingStepsManager
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
 from redbox.models import prompts
 from redbox.models.chat import DecisionEnum, ToolEnum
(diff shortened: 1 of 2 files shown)


redbox-core/redbox/models/chain.py
```diff
@@ -117,6 +117,31 @@ class Source(BaseModel):
         default="",
     )
 
+    @field_validator("document_name", mode="before")
+    def validate_document_name(cls, value):
+        if value == "" or value is None:
```
david-okeke1337, line 122: "not" is truthy/falsy so should be fine here, not super important but a bit more concise and inclusive
```suggestion
        if not value:
```

redbox-core/redbox/models/chain.py
```diff
@@ -117,6 +117,31 @@ class Source(BaseModel):
...
+            return cls.model_fields["document_name"].default
+        return value
+
+    @field_validator("source_type", mode="before")
+    def validate_source_type(cls, value):
+        if value == "" or value is None:
```
david-okeke1337, line 128:
```suggestion
        if not value:
```
Same as above

david-okeke1337 (approved):

PR #336, opened 2025-06-26
Fix None in Json

saisakul:
## Context

LLM returns None as value causing `ValidationError`. An equivalent of None in Json is null.

## What

- In parser if we detect JSON decoding error, then change from None to null
- Modify test case to contain invalid JSON (containing None)

- [ ] No


nora-errouhly:
testing shows token being streamed correctly

PR #339, opened 2025-06-30
Changing error message to be hidden

zainabmohamoud:
## Context

Changing error message to be hidden in chat to only appear in message.

## What

A more distinct error message following GDS styling

- [ ] No


nora-errouhly:
we agreed that the refresh issue also occur In dev so it is not related to this ticket. it would require creating a new ticket

PR #350, opened 2025-07-02
fix: redundant iteration

nboyse:
## Context

We have an iteration for attempting to connect to creds, this doesnt need to be there anymore - custom terraform mitigates this issue and it doesnt apply to redbox

## What

Removes legacy attempt to connect to transcribe service. We are covered for this.

- [X] No


ugo-daintta:
lgtm

PR #358, opened 2025-07-03
fix: prevent nonetype url error

nboyse:
## Context

We are getting an error when users access a chat in which the document has been deleted previously, there is no longer an associated url

## What

First check the file still exists, then create the URL

- [X] No

## Relevant links
https://sentry.ci.uktrade.digital/organizations/dit/issues/160898/?project=272&referrer=slack


ugo-daintta (approved):
Sprry just saw you updated. Approved

PR #373, opened 2025-07-07
add default page number

saisakul:
## Context

To prevent passing page_number = None to LLM, we specify the default value for page number.

## What

Use a default page number value if unstructured unable to find it.

- [ ] No

### DIFF
--- a/redbox-core/redbox/loader/loaders.py
+++ b/redbox-core/redbox/loader/loaders.py
@@ -153,7 +153,9 @@ class UnstructuredChunkLoader:
                 metadata=UploadedFileMetadata(
                     index=i,
                     uri=file_name,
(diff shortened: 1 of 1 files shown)


ugo-daintta (commented):
Might be cleaner to just say

```python
page_number =raw_chunk["metadata"].get("page_number", 1) 
```

saisakul:
> Might be cleaner to just say
> 
> ```python
> page_number =raw_chunk["metadata"].get("page_number", 1) 
> ```

The above code will use 1 if it cannot find page_number key, but in this case we have a page number with a value `None` which we want to replace with 1.

ugo-daintta (approved):

PR #405, opened 2025-07-23
Adding inclusive ingest process files

ugo-daintta:
## Context
Updating the file ingest process for tabular files (.csv, .xls, .xlsx).

## What
1. When a csv or excel file is uploaded, the file is read in to `pandas` to confirm it can be read in as a dataframe - this acts as a validation test
2. If validated, a single `chunk` is created with all of the text from the file(or sheet) stored in `chunk["text"]` - bypassing unstructured
3. This is then stored as a `tabular` document chunk in OpenSearch
4. If the document is an excel file, the sheet names are also stored in <table_name> html-like tags at the start of the chunk.

- [ ] No

1. Please do a code review
2. Test that the upload for csv and simple excel files works.
3. Run the notebook `tabular_opensearch_read.ipynb`. It will print out the last x created chunks (currently defaults to 6).
5. Confirm that it creates `normal`, `largest` and `tabular` chunk resolution chunks
6. Confirm that the text displayed in `tabular` chunks  looks like a csv i.e. with commas and new lines.
7. If the file is an excel file, it should start with a <table_name> tag which contains the table name.

## Relevant links
https://uktrade.atlassian.net/browse/REDBOX-990


nora-errouhly (approved):
Tested the app and reviewed the code. The tabular ingestion has been successfully added without impacting the rest of the ingestion.

PR #410, opened 2025-08-07
fix: include data dog vars in Dockerfiles

nboyse:
## Context

We need these so that data dog can link directly to the exact line of code that causes an error

## What

Adding vars that will help debug issues on data dog

- [X] No

## Relevant links
https://docs.datadoghq.com/integrations/guide/source-code-integration/?tab=python#embed-git-information-in-your-build-artifacts


ugo-daintta:
Builds successfully in my local 👍

ugo-daintta (approved):

PR #415, opened 2025-08-19
fix: replace deprecated unocov

nboyse:
## Context

Unocov is deprecated, need to use unoserver. This doesnt have a debian package, so we need to use poetry otherwise doc to docx wont work

## What

Remove unocov, replace with unoserver in poetry management

- [ ] No

Can you still upload a doc file?

## Relevant links
https://github.com/unoconv/unoconv?tab=readme-ov-file#unoconv-is-deprecated


ugo-daintta:
lgtm.

Local build is successful and docx upload completes successfully

PR #421, opened 2025-08-21
Feature/legislation route

richiebird:
## Context

As a Redbox@DBT user
I want to use @legislation as a route in my prompt
So that I can get answers that only originate from [Legislation.gov.uk](http://legislation.gov.uk/)

## What

Adds a route @legislation to invoke search legislation tool to make web search to http://legislation.gov.uk/

- [ ] No

Build locally, use prompts prefixed with @legislation to check that answers relate to the question asked, and citations only point to http://legislation.gov.uk/ and are valid urls

## Relevant links

https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-1012


nboyse (approved):
beautiful

PR #426, opened 2025-09-01
Bugfix/no tool call

saisakul:
## Context

Error when Redbox trying to do tool invocation when there is no tool calls.

https://uktrade.atlassian.net/browse/REDBOX-1058?atlOrigin=eyJpIjoiMjYzZjY4ZTZhMmNiNGY3NGI3MDNkN2RmNzcxYTZlNjQiLCJwIjoiaiJ9

## What

Fixing the code so that when there is no tool call, the worker agent returns its content.

- [ ] No

### DIFF
--- a/redbox-core/redbox/graph/nodes/sends.py
+++ b/redbox-core/redbox/graph/nodes/sends.py
@@ -69,12 +69,7 @@ def run_tools_parallel(ai_msg, tools, state, timeout=30):
     futures = []
     if len(ai_msg.tool_calls) == 0:
         # No tool calls
-        if type(ai_msg) is dict:
-            return ai_msg.AIMessage.content
-        elif type(ai_msg) is AIMessage:
-            return ai_msg.content
-        else:
-            log.error(f"The response is in an unexpected format {type(ai_msg)}")
+        return ai_msg.content
     else:
         try:
             # Use ThreadPoolExecutor for parallel execution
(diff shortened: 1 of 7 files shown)


redbox-core/redbox/graph/nodes/sends.py
```diff
@@ -67,40 +67,44 @@ def _tool_send(state: RedboxState) -> list[Send]:
...
-            return responses
-    except TimeoutError:
-        log.warning(f"Tool execution timed out after {timeout} seconds")
-        responses.append(AIMessage("Some tools timed out during execution"))
-    except Exception as e:
-        log.warning(f"Unexpected error in tool execution: {str(e)}", exc_info=True)
-        responses.append(AIMessage(f"Execution error: {str(e)}"))
+    if len(ai_msg.tool_calls) == 0:
+        # No tool calls
+        return ai_msg.content
```
nora-errouhly, line 72: are you returning the message from the worker agent (without the response from tool calls)? how is returning the message from worker agent impacting the final answer from evaluator agent? does the evaluator agent returns a good answer?

nora-errouhly:
There was an error when I tested your prompt.
AttributeError(\"'AIMessage' object has no attribute 'AIMessage'\"). I will copy the full log in JIRA and make suggestions

nora-errouhly (approved):

PR #449, opened 2025-09-11
Add web search agent

saisakul:
## Context

User needs to be able to access up-to-date information which is not available in LLM model. This PR creates a web search agent which has access to web search tool. It can search across websites or on specific domain.

## What

- create web search agent
- update root graph so that new route is the default route unless keyword is specified.
- update and add new route tests

- [ ] Yes (add why you have not)
- [ ] No

- [ ] No

### DIFF
--- a/redbox/tests/graph/test_multiagents_app.py
+++ b/redbox/tests/graph/test_multiagents_app.py
@@ -338,7 +338,6 @@ class TestNewRoutes:
             old_plan = [
                 ChainChatMessage(
                     role="ai",
-                    # text=escape_curly_brackets(MultiAgentPlan(tasks=tasks).model_dump_json()),
                     text=(MultiAgentPlan(tasks=tasks).model_dump_json()).replace("{", "{{").replace("}", "}}"),
                 )
             ]
--- a/redbox/tests/graph/test_app.py
+++ b/redbox/tests/graph/test_app.py
@@ -3,8 +3,8 @@ import logging
 
 # from enum import Enum
 from typing import Any
-from uuid import uuid4
 from unittest.mock import MagicMock
+from uuid import uuid4
 
 import pytest
 from langchain_core.documents import Document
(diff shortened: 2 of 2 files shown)


redbox/redbox/graph/root.py
```diff
@@ -146,28 +149,11 @@ def lm_choose_route_wrapper(state: RedboxState):
...
-        },
-    )
-    builder.add_conditional_edges(
-        "any_documents_selected",
-        documents_selected_conditional,
-        {
-            True: "is_new_route_enabled",
-            False: "chat_graph",
+            ChatRoute.chat: "chat_graph",
+            # ChatRoute.legislation: "legislation_graph",
```
sophie-daintta, line 153: Looks as though this might be a legacy or testing comment?

redbox/tests/graph/test_multiagents_app.py
```diff
@@ -0,0 +1,377 @@
...
+            tasks += [AgentTask(task="This is a fake task", agent=agents[i], expected_output="This is a fake output")]
+
+        planner = MultiAgentPlan(tasks=tasks)
+        old_plan = []
+        if "modify" in user_feedback:
+            # old plan here
+            old_plan = [
+                ChainChatMessage(
+                    role="ai",
+                    # text=escape_curly_brackets(MultiAgentPlan(tasks=tasks).model_dump_json()),
```
sophie-daintta, line 341: Looks like a legacy comment I think.

redbox/tests/graph/test_multiagents_app.py
```diff
@@ -0,0 +1,377 @@
...
+                [3, 10_000],
+                True,
+                AgentEnum.Internal_Retrieval_Agent,
+                ANSWER_WITH_CITATION,
+            ),
+        ],
+    )
+    @pytest.mark.asyncio
+    async def test_newroute_zero_or_one_task(
+        self, test_name, user_prompt, documents, has_task, agent, evaluator, mocker: MockerFixture
```
sophie-daintta, line 269: Not that it changes much but I think test_name isn't used in this function - is it useful to keep for clarity though?

redbox/redbox/graph/root.py
saisakul: We need to fix the legislation graph (in different PR) so I commented this out so that it is not accessible by user.

sophie-daintta (approved):
Looks good to me! Nice refactoring on the multiagent tests.

PR #459, opened 2025-09-17
Feature/tabular worker

nora-errouhly:
integrating tabular into new route.
built new tabular agent as part of newroute
see architecture of new route (pls note that the graph display is not exact due to send_tasks, e.g planner)
image

how to test:
- select any tabular file in UI. ask question without keyword to trigger newroute


nora-errouhly:
test example shown here
Screenshot 2025-09-17 at 16 25 04

nboyse (approved):

PR #481, opened 2025-10-07
added monitoring of web search agent queries and results

richiebird:
## Context

I want to collect information on what information is sent to Kagi, so that we can 1. monitor if sensitive information are being sent to Kagi and 2. identify websites that are regularly searched by user so that we can inform future decision on knowledge base.

## What

Added a table to the django admin that records user query, user message id, selected files for that message (in a separate table) and the website urls returned by agents that invoke kagi web search calls

- [ ] Yes
Unit tests already cover the kagi web search tool

- [ ] No

Will require tester to update kagi api key in their .env file to latest.

Deploy redbox locally, and test some example queries that will call web search. Check the "monitor web search results" tab on the django admin page and see if queries are accurately recorded. If a document is selected as part of the prompt this should also be displayed when clicking on a web search query record.

Examples: 
- What is the latest temporary piece of legislation on road traffic in Scotland?
- When is the release date for the book the doors of stone?

Because multiple agents can do a web search, also test a complex prompt that invokes both web search agent and legislation agent (e.g. combine the above examples into 1 single prompt) and check that the "monitor web search results" tab on the django admin page covers all web urls (including legislation.gov.uk specific urls).

## Relevant links
https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-1074


django_app/redbox_app/redbox_core/admin.py
```diff
@@ -308,6 +308,17 @@ class MonitorSearchRouteAdmin(admin.ModelAdmin):
...
 
 
+class MonitorWebSearchResultsAdmin(admin.ModelAdmin):
+    list_display = [
+        "chat_message",
+        "user_text",
+        "web_search_urls",
+        "created_at",
+    ]
+    readonly_fields = ["selected_files"]
```
sophie-daintta, line 318: Just wondering why this is the only read-only field, where the rest can be altered?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -945,6 +945,16 @@ def __str__(self):
...
 
 
+class MonitorWebSearchResults(UUIDPrimaryKeyBase, TimeStampedModel):
+    chat_message = models.ForeignKey(ChatMessage, on_delete=models.CASCADE)
+    user_text = models.TextField(max_length=32768, null=False, blank=False)
+    selected_files = models.ManyToManyField(File, related_name="+", symmetrical=False, null=True, blank=False)
+    web_search_urls = models.TextField(max_length=32768, null=False, blank=False)
+
+    def __str__(self):
+        return f"{self.user_text} {self.web_search_urls}"
```
sophie-daintta, line 955: Just for pure appearance reasons in the django admin form, this ends up being a very long title on individual forms. Would it make sense to just have the user text?

sophie-daintta (approved):
Looks good.

PR #516, opened 2025-10-22
Feature/tabular document citations

sophie-daintta:
## Context

When running a query that requires document and citations sources eg. A query which requires the web search agent and the tabular agent, this PR allows both types of source to be included. 

## What

This PR contains:

- A fix allowing both document and citation sources.
- A small fix to the tabular agent, breaking the loop when a SQL query is not generated. 
- A fix to the test for the tabular agent adding a dummy SQL query.

- [x] Yes

- [ ] No

Run tests which should require multiple agents and check that citation and document sources can be returned.

### DIFF
--- a/redbox/tests/graph/test_multiagents_app.py
+++ b/redbox/tests/graph/test_multiagents_app.py
@@ -185,7 +185,10 @@ WORKER_RESPONSE = AIMessage(
         "tool_calls": [
             {
                 "id": "call_e4003b",
-                "function": {"arguments": '{\n  "query": "ai"\n}', "name": "_some_tool"},
+                "function": {
+                    "arguments": '{\n  "query": "ai"\n, \n  "sql_query": "SELECT 1;"\n}',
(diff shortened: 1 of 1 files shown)


redbox/redbox/graph/nodes/processes.py
```diff
@@ -796,15 +796,17 @@ def _build_tabular_agent(state: RedboxState):
...
             messages.append(AIMessage(ai_msg["messages"][-1].content))
             try:
                 messages.append(
                     AIMessage(f"Here is the SQL query: {ai_msg["messages"][-1].tool_calls[-1]['args']['sql_query']}")
                 )
             except Exception:
                 log.info("no sql query input to tool")
-
-            num_iter += 1
+                sql_error = ai_msg["messages"][-1].content
```
nora-errouhly: since there is no sql query input to the tool, there is not execution to the database and therefore No sql error is returned. This line need to be removed as the message content does not represent a SQL error

PR #533, opened 2025-10-30
switch to and added brave web search functionality

richiebird:
## Context

Decision to switch to Brave as our web search api provider

## What

Changed settings default value to Brave, adjusted web search call tool to process both Brave and Kagi web search responses and create documents based on their respective json outputs

- [X] Yes
added unit tests for specific handling of Brave and Kagi responses to document mapping

- [ ] No
Check web search on new route works as intended, and is making calls to the Brave API and not the Kagi API

## Relevant links
https://uktrade.atlassian.net/jira/software/projects/REDBOX/boards/558?selectedIssue=REDBOX-1169

### DIFF
--- a/redbox/tests/test_tools.py
+++ b/redbox/tests/test_tools.py
@@ -23,7 +23,7 @@ from redbox.graph.nodes.tools import (
 )
 from redbox.models.chain import AISettings, RedboxQuery, RedboxState
 from redbox.models.file import ChunkCreatorType, ChunkMetadata, ChunkResolution
-from redbox.models.settings import Settings, get_settings
+from redbox.models.settings import Settings
 from redbox.test.data import RedboxChatTestCase
 from redbox.transform import bedrock_tokeniser, combine_documents, flatten_document_state
 from tests.retriever.test_retriever import TEST_CHAIN_PARAMETERS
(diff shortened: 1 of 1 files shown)


redbox/redbox/graph/nodes/tools.py
```diff
@@ -430,35 +447,74 @@ def web_search_with_retry(query: str, no_search_result: int = 20, max_retries: i
...
-def kagi_search_call(query: str, no_search_result: int = 20) -> tool:
+def kagi_response_to_documents(response: requests.Response) -> list[Document]:
+    tokeniser = bedrock_tokeniser
+    mapped_documents = []
+    results = response.json().get("data", [])
+    for i, doc in enumerate(results):
+        # extract only search results asper kagi documentation
+        if doc["t"] == 0:
+            page_content = "".join(doc.get("snippet", []))
+            token_count = tokeniser(page_content)
```
saisakul: can we refactor this bit so we don't duplicate the codes?

redbox/tests/test_tools.py
```diff
@@ -283,6 +301,80 @@ def test_gov_tool_params():
     assert len(documents) == ai_setting.tool_govuk_returned_results
 
 
+@requests_mock.Mocker(kw="mock")
+def test_kagi_response_to_document(**kwargs):
+    env = get_settings()
+    kwargs["mock"].get(
+        env.web_search_settings().end_point,
```
saisakul: this is incorrect. This will always be whatever the provider was set to, which is not necessary Kagi. Please mock it for kagi

PR #606, opened 2025-12-01
fix: fixes bug in _build_agent that throws away tool responses over max tokens

hwixley:
## Context
InternalRetrievalAgent would return no response if it's tool responses were over the max tokens.

## What
- Fix: Updates build_agent to truncate responses over max tokens rather than not use it
- Fix: Updates chunk resolution filter to use enum value for string serialisation
- Adds further logging for better observability in deployments

- [x] Yes

Added basic unit test for truncation function `truncate_to_tokens` in `transform.py`

- [ ] No

Quite hard to test as the size of tool outputs is non-deterministic. Requires you to deploy the application and watch the logs to see if a response is over max tokens, if-so verify the response is still received on the frontend (after being truncated).

### DIFF
--- a/redbox/redbox/retriever/retrievers.py
+++ b/redbox/redbox/retriever/retrievers.py
@@ -152,7 +152,7 @@ def query_to_documents(
         hits = response.get("hits", {}).get("hits", [])
         documents = [hit_to_doc(hit) for hit in hits]
         logger.warning("[query_to_documents] Query returned %d documents", len(documents))
-        logger.warning("Documents: %s", documents)
+        logger.warning("[query_to_documents] Documents: %s", documents)
         return documents
 
     except OpenSearchException as e:


redbox/redbox/retriever/retrievers.py
```diff
@@ -86,11 +86,14 @@ def _get_relevant_documents(self, query: str, *, run_manager: CallbackManagerFor
...
-        logger.info("query to opensearch: from get_relevant_documents")
-        logger.info(str(body))
-        response = self.es_client.search(index=self.index_name, body=body)
-        logger.info("response from opensearch: from get_relevant_documents")
-        logger.info(str([self.document_mapper(hit) for hit in response["hits"]["hits"]]))
+        logger.warning("query to opensearch: from get_relevant_documents")
+        logger.warning(str(body))
+        try:
+            response = self.es_client.search(index=self.index_name, body=body)
+        except Exception as e:
```
dougmills-DIT: Not logged and not raised, plus generic exception handling not ideal - can we be more specific?

PR #618, opened 2025-12-03
Rename 'Golden Rules' Document Reference

saisakul:
## Context

As a user of the Submissions Checker tool, I want clear, professional document naming so that I understand what guidance is being referenced.

## What

- updated wording in submission prompt from `golden rules` to `Ministerial Submission Template Guidance`
- revised prompt to ensure the specific wording is used for reference

- [ ] Yes

- [x] No


redbox/redbox/models/prompts.py
```diff
@@ -501,4 +501,5 @@
 - AVERAGE SCORE: A simple mean of the score across all 7 criteria.
 - ASSESSMENT SUMMARY: A brief statement of the overall quality of the submission. Be critical but constructive in your feedback.
 - Make reference or citations to the knowledge base information.
+    - when referenceing to template guidance, references should consistently use ‘Ministerial Submission Template Guidance’
```
david-okeke1337: "referencing" ⁠ 🤓

PR #656, opened 2025-12-11
feat: setup per tool timeouts in parallel executor

hwixley:
## Context
So each tool in parallel executor has it's own max timeout to catch any throttling tools.

## What
- Sets up generic 30s per-tool timeout in `run_tools_parallel` as a new func param
- Adds a unique run_id UUID to `run_tools_parallel` logs so that tool logs in CloudWatch indicate which parallel tool run they originate from

- [ ] Yes

- [ ] No

1. Do chat which requires tool execution, ie. "search the web for..."
2. Validate tools execute and return results
3. Validate logs display correctly


redbox/tests/graph/nodes/test_sends.py
```diff
@@ -171,7 +170,7 @@ def test_tool_time_out(self, mocker: MockerFixture):
 
         response = run_tools_parallel(ai_msg=ai_msg, tools=[search_wikipedia], state=fake_state)
 
-        assert response == []
+        assert response is None
 
     @pytest.mark.parametrize("side_effect", [(TimeoutError("Thread time out")), (Exception("Thread error"))])
     def test_threadpool_time_out(self, side_effect, mocker: MockerFixture):
```
saisakul: please add test case when 1 tool completes the request and 1 tool timeout.

redbox/redbox/graph/nodes/sends.py
```diff
@@ -64,71 +66,122 @@ def _tool_send(state: RedboxState) -> list[Send]:
...
+    thread.daemon = True  # The thread will exit when the main program exits
+    thread.start()
+    thread.join(timeout)  # applying timeout constraint
+
+    if not completed[0]:  # if it times out
+        log.warning(f"Tool execution timed out after {timeout} seconds")
+        return None
+    if exception[0]:  # if the tool fails
+        log.warning(f"Tool execution failed: {str(exception[0])}")
+        return None
```
hwixley: should we propagate the exception to the caller `run_tools_parallel` for handling ? otherwise exceptions raised in tool calls may be silent

nora-errouhly: thanks for the input. I have added your suggestion

saisakul (approved):

PR #658, opened 2025-12-11
fix: add max token truncation logic to agent_with_loop

hwixley:
## Context
To match truncation logic in `build_agent` - this truncation logic fixes a bug where if tool responses are over max tokens they are unused instead of truncated resulting in empty responses

## What
Ensures tool responses are truncated to fit max_tokens if they go over rather than discarding the results

- [ ] Yes

- [ ] No

1. Do a chat request that results in high token output tool responses, this can be replicated when doing heavy operations on documents, for example multiple operations on large documents
2. Validate from logs that a tool response went over max tokens and if-so validate it was still used


redbox/redbox/graph/nodes/processes.py
```diff
@@ -645,23 +655,48 @@ def local_loop_condition():
...
+                        if remaining_tokens > 0:
+                            log.warning(
+                                f"{log_stub} Truncating tool output to fit remaining token budget ({remaining_tokens})."
+                            )
+                            truncated = truncate_to_tokens(res.content, remaining_tokens)
+                            result_content.append(truncated)
+                            current_token_counts += bedrock_tokeniser(truncated)
+                        else:
+                            log.warning(f"{log_stub} No remaining token budget ({max_tokens}). Skipping.")
+                        break  # Max reached — stop processing further results
```
dougmills-DIT: Unit tests? I had a look and i can't see any existing tests covering this code

dougmills-DIT (changes requested):

PR #740, opened 2026-01-16
fix: codepipeline integration with Procfile

nboyse:
- [ ] Yes

- [ ] No


nboyse:
fixed by https://github.com/uktrade/redbox-deploy/pull/90

PR #804, opened 2026-02-06
add artifact agent

saisakul:

## Context

Adding a new Artefact Builder agent which can take any format, and criteria for creating an aritifact.

## What

- add new `ArtifactAgent` which extends `WorkerAgent` but have update `artefact_criteria` property rather agents_results`.
- add new Artefact Builder agent with its config 
- update the retriever so it default to return 30 files metadata
- add artefact_criteria in state so it can be used by Evaluator agent 
- update the `build_retrieve_knowledge_base` tool to allow extraction of a specific file
- 

- [x] Yes

- [x] No

### DIFF
--- a/redbox/tests/graph/agents/test_agent.py
+++ b/redbox/tests/graph/agents/test_agent.py
@@ -2,7 +2,8 @@ import pytest
 from langchain_core.messages import AIMessage
 from pytest_mock import MockerFixture
 
-from redbox.graph.agents.configs import AgentConfig, PromptConfig, PromptVariable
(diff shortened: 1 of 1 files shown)


redbox/redbox/retriever/queries.py
```diff
@@ -122,6 +123,7 @@ def get_minimum_metadata(
     )
 
     return {
+        "size": 30,
```
hwixley: If limiting to 30 should we sort by created date ?

PR #884, opened 2026-03-10
Feature/datahub feedback queue

nora-errouhly:
Currently, we have a Datahub agent that can retrieve company information. In case of multiple records, the Datahub agent should ask the user for feedback to confirm which company is relevant. Then Datahub agent should answer user question and any follow up question based on the chosen relevant company. If only a single record is returned from Datahub then no user feedback is requested.
ways to test:
- Extract companies details about cobalt. (should return multiple companies asking for user feedback).
- Extract companies details about Cobalt Crescent Holdings. (should not trigger user feedback)


hwixley:
Continuing on noras work:

- Added extra `metadata` field to server-side responses allowing the server to tell us if user feedback is required or not. If it is, a reason is provided and injected into agent prompt.
- Added new task status `REQUIRES_USER_FEEDBACK`
- Setup node in the graph that exits early if task has user feedback status and routes appropriately, allowing future tasks to use the same logic for user feedback
- Updated run_tools_parallel to parse the MCP metadata responses
- Forked build_agent_with_loop for datahub agent to avoid changing logic of loop agent - this is temporary given saisakul mentioned we want to refactor that to align with WorkerAgent class

nora-errouhly (commented):
new changes from Harry looks good. cannot approve PR myself as I posted the original PR.

david-okeke1337 (approved):

PR #889, opened 2026-03-11
bugfix: fix mcp server stale connections, and datahub agent tool_call hallucinations

hwixley:
## Context
Solves issues with app when there are connection issues with MCP server:

1. When app starts up and cant connect to MCP datahub agent has no tools, when MCP server starts up later app does not try reload the tools.

2. When unable to connect to the MCP server, Datahub agent has no tools. When this happens the agent is prone to start hallucinating tool calls in its message meaning it returns bad responses to user (ie. saying it is doing a search in background, or that the search failed).

## What
1. To prevent stale MCP connections, ensure the app retries to reload the MCP tools if not set using async task in consumer.

2. To prevent tool_call hallucinations being fed back to the user we added a check if tools is non-empty in `build_agent_with_loop` and only then call tool runner and successfully complete the task. Otherwise fail the task.

- [ ] Yes

- [ ] No

1. Run app without MCP server running
2. Check responses to datahub questions say it is unable to retrieve information
3. Start MCP server
4. App should reload datahub agent tools and be able to answer datahub questions


django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -69,6 +71,49 @@
 logger.info("WEBSOCKET_SCHEME is: %s", settings.WEBSOCKET_SCHEME)
 
 
+async def wait_for_health(
```
nora-errouhly: can you add docstrings to your new functions outlining what they are doing, and what they returns..

nora-errouhly (commented):
need to add more docstrings and comments especially on consumers.py

PR #931, opened 2026-03-27
fix: refactor run tools parallel to class

hwixley:
## Context
Refactor run_tools_parallel to dedicated Class for modular logic, allowing better testing, readability, and extensibility.

## What
Refactored logic in `run_tools_parallel` to class for more modular components, better testing capability, and better extensibility through more structured return types.

Flow:
1. Initialise runner ie. `my_runner = ToolRunner(...)`
2. Run the runner ie. `my_runner.run(tool_calls=...)`
3. Get results from structured return type dataclass `ToolExecutionResult` which includes responses and failed_tools

- [x] Yes

- [ ] No

Test the tool runner across all different agents/tools


lizzie-perry-dbt (approved):
Tested locally and agents work as expected for me! Nice work :)

redbox/redbox/graph/nodes/runner/runner.py
```diff
@@ -0,0 +1,234 @@
...
+from redbox.graph.nodes.runner.wrap_async import wrap_async_tool
+
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class ToolExecutionResult:
+    """Result of parallel tool execution."""
+
+    responses: List[AIMessage]
```
saisakul: please add in field description and default value.

redbox/redbox/graph/nodes/runner/runner.py
```diff
@@ -0,0 +1,234 @@
...
+    failed_tools: List[str]
+
+    @property
+    def is_success(self) -> bool:
```
saisakul: where are these properties used?

redbox/redbox/graph/nodes/sends.py
```diff
@@ -74,282 +66,39 @@ def _tool_send(state: RedboxState) -> list[Send]:
...
-                    f"{log_stub} Completed. Successful parallel tool responses: {len(responses)}. Responses: {responses}"
-                )
-                return responses
-            else:
-                log.warning(
-                    f"{log_stub} Every tool execution has failed or timed out after {per_tool_timeout} seconds."
-                )
-                return None
+        try:
+            result = runner.run(tool_calls=ai_msg.tool_calls)
```
saisakul, line 92: It seems we are not using `failed_tools` anywhere?

redbox/redbox/graph/nodes/runner/runner.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def _submit_all(self, tool_calls: list[ToolCall]) -> dict[Future, dict]:
+        """Submit every tool call to the executor, skipping and logging any that fail to launch."""
+        futures = {}
+
+        for tool_call in tool_calls:
+            tool_name = tool_call.get("name")
+            try:
+                res = self.submit(tool_call=tool_call)
+                if res is None:
```
saisakul: in what case do we expect to see res is None?

saisakul:
in `tool_exceptions.` tests, please assert that the content contains what was expected e.g.  tool x returned None

redbox/redbox/graph/nodes/runner/runner.py
hwixley: good spot, yes these are deprecated now will remove

redbox/redbox/graph/nodes/runner/runner.py
hwixley: yeah good spot, think this check was leftover from previous implementation that returned Nones in run_with_timeout (which is deprecated in favour of leveraging per-tool timeouts on future.result(...))

have removed this check and fixed self.submit signature to not have optional return type

redbox/redbox/graph/nodes/sends.py
hwixley: yes good spot, i am thinking of maybe doing a separate PR to leverage failure results

saisakul (approved):

PR #978, opened 2026-04-15
fix: fix datahub loop result collation

hwixley:
## Context
Fix datahub loop result collation - previously multiple results were not collated properly causing tool results to be lost

## What
- Fixes datahub tool result collation
- Adds better exception handling on wrap_async
- Deprecates wrap_async event loop to leverage existing one in parent caller run_with_timeout

- [x] Yes

- [ ] No


dougmills-DIT (commented):
Could you add unit tests to prove the bug is fixed?

redbox/redbox/graph/nodes/processes.py
```diff
@@ -699,37 +699,49 @@ def local_loop_condition():
...
+                    success = current_result[1]
+                    is_intermediate_step = eval(current_result[2])
+
+                    if len(current_result) > 3:
+                        reason = current_result[3]
+                        feedback_reasons.append(f"Failure reason: {reason}.\n\n{result_content}")
+                    else:
+                        collated_result += f"<tool_result_{i}>{result_content}</tool_result_{i}>"
+
+                    if success == "fail":
```
lizzie-perry-dbt, line 719: Is this not bool? Should it be? Unsure if there is a non pass/fail hence why String

redbox/redbox/graph/nodes/processes.py
```diff
@@ -699,37 +699,49 @@ def local_loop_condition():
...
-                result_content = result[0]
-                success = result[1]
-                is_intermediate_step = eval(result[2])
+                collated_result = ""
+                feedback_reasons = []
+                for i, r in enumerate(result):
+                    current_result = r.content  # this is a tuple
+                    # format of result: (result, success, is_intermediate_step)
+                    log.warning("my-overall-result")
+                    log.warning(current_result)
```
lizzie-perry-dbt: Maybe unneeded as warning but unsure and super minor
```suggestion
                    log.debug("my-overall-result")
                    log.debug(current_result)
```

lizzie-perry-dbt (approved):
Looks good :)

redbox/redbox/graph/nodes/processes.py
hwixley: yeah this is supposed to be string - this follows the same logic as build_agent_with_loop for datahub agent

think the initial reason it was set to string so could be put in AIMessage (which is what run_tools_parallel returns), however, in future want to refactor run_tools_parallel so can leverage more structured return types so dont need these poorly typed strings

dougmills-DIT (approved):

PR #999, opened 2026-04-24
Add callback for token usage

saisakul:
## Context

We need to provide token usage to Datadog so that it can calculate the LLM cost.

## What

- Add a callback to add annotation to the LLM span
- Change from auto instrument to manual instrument. Need to remove `botocore` from auto otherwise it will overwrite the token usage that we annotated.  
- Add tests for new callback function

- [x] Yes
- [] No (add why you have not)

tested locally and in dev

- [ ] No

run some prompts in dev and then check that DataDog traces calculate token costs

### DIFF
--- a/redbox/redbox/api/callbacks.py
+++ b/redbox/redbox/api/callbacks.py
@@ -37,13 +37,22 @@ class LoggerCallbackHandler(BaseCallbackHandler):
 
 
 class TokenAnnotationCallback(BaseCallbackHandler):
+    def _extract_model_name(self, response_metadata) -> str:
+        model_name = response_metadata.get("model_name", None)
+        if model_name:
+            if model_name.startswith("arn"):
+                # grab the last bit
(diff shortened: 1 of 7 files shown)


redbox/redbox/api/callbacks.py
```diff
@@ -32,3 +34,31 @@ def on_chain_error(self, error: BaseException, **kwargs: Any):  # noqa:ARG002
...
+        model_name = response_metadata.get("model_name", None)
+        if model_name:
+            if model_name.startswith("arn"):
+                # grab the last bit
+                model_name = model_name.split("/")[-1]
+            model_name = model_name.split(".")[-1]
+        return model_name
+
+    def on_llm_end(self, response: LLMResult, **kwargs) -> None:
+        if response.generations[0]:
```
hwixley, line 41: NIT: may be worth checking list is non-empty before checking first element, ie.

`if respoonse.generations and response.generations[0]`

PR #1009, opened 2026-04-28
chore: add agent prompt config variable tests

hwixley:
## Context
A bug in datahub agent which was not caught by tests was missing variables in prompt config. This solves that issue for future.

## What
Automatically tests agent prompt configs 

- [x] Yes

- [ ] No

### DIFF
--- a/redbox/tests/graph/agents/test_agent.py
+++ b/redbox/tests/graph/agents/test_agent.py
@@ -2,7 +2,7 @@ import pytest
 from langchain_core.messages import AIMessage
 from pytest_mock import MockerFixture
 
-from redbox.graph.agents.configs import AgentConfig, PromptConfig, PromptVariable, agent_configs
+from redbox.graph.agents.configs import AgentConfig, PromptConfig, PromptVariable, agent_configs, prompt_configs
 from redbox.graph.agents.formats import ArtifactAgent
 from redbox.graph.agents.workers import WorkerAgent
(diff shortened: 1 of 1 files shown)


redbox/tests/graph/agents/test_agent.py
```diff
@@ -119,3 +119,18 @@ def test_post_processing(self, result, fake_state_with_plan):
...
         assert response["artifact_criteria"] == "A result"
         assert response["agent_plans"].get_task_status(task.id) == TaskStatus.COMPLETED
+
+
+@pytest.mark.parametrize(
+    "name, prompt_cfg",
+    [
+        (key, value)
+        for key, value in prompt_configs.items()
+        if key not in ["Planner_Agent", "Replanner_Agent", "Summarisation_Agent", "Evaluator_Agent"]
```
lizzie-perry-dbt, line 129: Do we need to add Datahub_Agent to this list as it also doesn't contain the tag in the DATAHUB_PROMPT ?
I assume this then could cause error - or maybe is not relevant for this?

lizzie-perry-dbt (approved):
One potential error with DataHub agent not containing the tag, so may need to be added to the list :)

redbox/tests/graph/agents/test_agent.py
hwixley: yeah the only agents included here are special cases, the datahub agent does need task and expected_output vars (which are in system prompt via prompts.DATAHUB_QUESTION_PROMPT)

PR #1013, opened 2026-04-29
fix: loop agent result handling

hwixley:
## Context
Bug in agent_with_loop not returning proper response object causing exceptions

## What
- Add check in build_agent_with_loop for None result types
- Fix bug in response handling in run_tools_parallel

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/redbox/graph/nodes/processes.py
+++ b/redbox/redbox/graph/nodes/processes.py
@@ -540,7 +540,10 @@ def build_agent_with_loop(
             log.warning(f"{log_stub} Running tools via run_tools_parallel...")
             result = run_tools_parallel(ai_msg, tools, state, is_loop=True)  # this agent runs with loop
 
-            if has_loop and len(ai_msg.tool_calls) > 0:  # if loop, we need to transform results
+            if not result:
+                log.error("[build_agent_with_loop] Result from run_tools_parallel is None")
+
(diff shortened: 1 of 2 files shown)


redbox/redbox/graph/nodes/processes.py
```diff
@@ -540,7 +540,10 @@ def local_loop_condition():
             log.warning(f"{log_stub} Running tools via run_tools_parallel...")
             result = run_tools_parallel(ai_msg, tools, state, is_loop=True)  # this agent runs with loop
 
-            if has_loop and len(ai_msg.tool_calls) > 0:  # if loop, we need to transform results
+            if not result:
+                log.error("[build_agent_with_loop] Result from run_tools_parallel is None")
+
```
lizzie-perry-dbt, line 545: Would adding a continue after log error make more stable? as would always skip rest of the iteration

lizzie-perry-dbt (approved):
Minor comment for potential fragile case, but don't think is s definite, so happy to merge :)

redbox/redbox/graph/nodes/processes.py
hwixley: yeah good idea, only reason did not do that is given we currently actually update response to say "no results returned" in the case of failure so the model knows that a tool failed

PR #1020, opened 2026-04-30
Added datahub agent to normal assist, along with test for graph

lizzie-perry-dbt:
## What

Add Datahub_agent to normal assist

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/tests/graph/test_root.py
+++ b/redbox/tests/graph/test_root.py
@@ -58,8 +58,8 @@ class TestNewRouteGraphs:
 
             assert "Fake_Agent" not in graph.nodes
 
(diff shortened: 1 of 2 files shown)


redbox/tests/graph/test_root.py
```diff
@@ -56,3 +57,9 @@ def test_non_existent_node(self):
             ).get_graph()
 
             assert "Fake_Agent" not in graph.nodes
+
+    def test_datahub_agent_prompt_variables():
```
hwixley: think you just need to add `self` to args given in test class

hwixley (approved):
Nice! Just need to fix failing test and you should be good

PR #1030, opened 2026-05-12
fix when no text to embed

saisakul:
## Context

Bugfix: In gov.uk search tool, embedding fails when there is no input. 

## What

- if there is no text, don't do embedding and assign similarity score to 0 (not similar)

- [ ] Yes
add test case to existing test

- [x] No


redbox/redbox/graph/nodes/tools.py
```diff
@@ -499,11 +499,14 @@ def recalculate_similarity(response, query, num_results):
         embedding_model = get_embeddings(get_settings())
         em_query = embedding_model.embed_query(query)
         for r in response.get("results"):
-            text_compare = r.get("description") if r.get("description") else r.get("indexable_content")[:500]
-            em_des = embedding_model.embed_query(text_compare)
-            r["similarity"] = cosine_similarity(np.array(em_query).reshape(1, -1), np.array(em_des).reshape(1, -1))[0][
-                0
-            ]
+            text_compare = r.get("description", "") if r.get("description") else r.get("indexable_content", "")[:500]
```
david-okeke1337: Can we simplify this a bit
```suggestion
            text_compare = r.get("description") or r.get("indexable_content", "")[:500]
```

david-okeke1337 (approved):

PR #1031, opened 2026-05-12
Bug: Input is too long for requested model in Evaluator Agent

lizzie-perry-dbt:
## Context

Error generated from sending too long input to Claude model in Evaluator agent.

At the present the evaluator only accounts for the size of previous chat history, however it does not account for the size of the inputs from other agents, therefore the input size could exceed the context limit.

## What

Implement the fix so that we post process information sent from upstream agents if the context size is exceeded.
Before: the input size check only counted the chat history, and not the agents output, so the total prompt could exceed the context limit and error.
What was done:
- updated combinre_agents_state to truncate each agent's output if it is too big, with a [truncated] marker so the model knows that the output was truncated / is incomplete
- updated evaluator to calculate the size available remaining and pass that on as the budget
- trimmed the joined result so rounding and markers doesn't push it over the limit 

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/redbox/chains/runnables.py
+++ b/redbox/redbox/chains/runnables.py
@@ -1,7 +1,7 @@
 import logging
 import re
-from typing import Any, Callable, Iterable, Iterator
 from datetime import date
+from typing import Any, Callable, Iterable, Iterator
 
 from langchain_core.callbacks.manager import CallbackManagerForLLMRun, dispatch_custom_event
 from langchain_core.language_models import BaseChatModel
@@ -25,7 +25,7 @@ from redbox.models.chain import ChainChatMessage, PromptSet, RedboxState, get_pr
 from redbox.models.errors import QuestionLengthError
 from redbox.models.graph import RedboxEventType
 from redbox.models.settings import ChatLLMBackend, get_settings
-from redbox.transform import bedrock_tokeniser, flatten_document_state, get_all_metadata
+from redbox.transform import bedrock_tokeniser, combine_agents_state, flatten_document_state, get_all_metadata
 
 log = logging.getLogger()
 re_string_pattern = re.compile(r"(\S+)")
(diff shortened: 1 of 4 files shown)


redbox/redbox/chains/runnables.py
```diff
@@ -104,7 +104,35 @@ def _chat_prompt_from_messages(state: RedboxState) -> Runnable:
...
+        # Agent results take priority over chat history
+        # Size the agent results first against the available budget, then chat history fits into the remainder
+        history_reservation = prompts_budget
+
+        if "agents_results" in _additional_variables:
+            other_vars_token = _tokeniser(
+                (_additional_variables.get("artifact_criteria") or "") + _additional_variables.get("todays_date", "")
+            )
+
+            safety_margin = 2000
```
hwixley: what is this safety margin for ?

redbox/tests/test_transform.py
```diff
@@ -481,3 +483,45 @@ def test_join_result_with_token_limit(
...
+            "multiple-medium-agent-outputs--aggregate-exceeds-context-window",
+            {f"Agent_{i}": AIMessage(" ".join(["filler"] * 15000)) for i in range(10)},
+        ),
+        (
+            "mixed-small-and-huge-agent-outputs--combined-exceeds-context-window",
+            {"Small_Agent": AIMessage("Short result"), "Huge_Agent": AIMessage(" ".join(["x"] * 200000))},
+        ),
+    ],
+)
+def test_combine_agents_state_fits_within_eval_agent_max(test_name, agents_results):
```
hwixley, line 512: could be worth adding some tests around the presence of truncation character

redbox/redbox/transform.py
```diff
@@ -342,9 +342,46 @@ def max_score(group: list[Document]) -> float:
...
+
+    if total_tokens <= max_tokens:
+        flatten_agent_results = "\n\n".join([msg.content for msg in agents_results.values()])
+        return {"all_result": flatten_agent_results}
+
+    marker_cost = bedrock_tokeniser(TRUNCATION_MARKER)
+    parts: list[str] = []
+    for agent_id, msg in agents_results.items():
+        original_tokens = sizes[agent_id]
+        share = max(1, int(max_tokens * original_tokens / total_tokens))
```
hwixley: maybe worth to do `math.floor` over `int` in case of rounding up issues ?

redbox/redbox/transform.py
```diff
@@ -342,9 +342,46 @@ def max_score(group: list[Document]) -> float:
...
+    Combine a list of agent results into a string.
+    OR if it is over the max amount, truncating proportially
+    A truncation marker is appended to any block, so the downstream LLM knows that the content is incomplete
+    """
     if not agents_results:
         return {}
-    flatten_agent_results = "\n\n".join([msg.content for msg in agents_results.values()])
-    return {"all_result": flatten_agent_results}
+    sizes = {aid: bedrock_tokeniser(msg.content) for aid, msg in agents_results.items()}
+    total_tokens = sum(sizes.values())
```
hwixley, line 359: NIT: think this is a bit of an edge case but given we do a "\n\n".join() on results, is it worth adding a calculation to incorporate token counts of joining with those "\n\n" chars ?

hwixley (commented):
Nice stuff! love the 'share-based' token quota logic - just left a few comments

redbox/redbox/chains/runnables.py
lizzie-perry-dbt: It is a buffer for extra tokens like formatting, separators, metadata etc. added by the template, or small differences between the bedrock tokeniser and the model's tokeniser. I will put this in a comment

redbox/tests/test_transform.py
lizzie-perry-dbt: Good shout, tests added

redbox/redbox/transform.py
lizzie-perry-dbt: Yeah this is a good point. I think it should be ok as the check at the bottom of the function re-truncated the joined string if it goes over the limit, so the safety net should handle this. But could take away up front

hwixley (approved):

PR #1039, opened 2026-05-19
feat: add new mcp result type for citations on paged 'composite' results

hwixley:
## Context
To facilitate citations on new exploratory company activity tool we need to be able to handle 'paged composite' results. 'Composite' result types are already supported and are used for results which include information about a parent datapoint and respective child datapoints.

To support exploratory querying across multiple companies we need to support paged versions of these composite results.

## What
This PR adds support for formatting and parsing for paged composite types allowing citations across all datapoints.

- [x] Yes

- [ ] No

- This is currently only supported via new exploratory company activity MCP tool https://github.com/uktrade/data_hub_mcp/pull/66

### DIFF
--- a/redbox/redbox/api/format.py
+++ b/redbox/redbox/api/format.py
@@ -85,6 +85,13 @@ def format_mcp_tool_response(tool_response, creator_type: ChunkCreatorType) -> t
             paged_results = [v for v in paged_data.values() if v is not None]
             for page in paged_results:
                 deep_links += [(item.get("url"), item) for item in page.get("result", {}).get("items", [])]
+        case "paged_composite":
+            for company, paged_activity_data in result.get("items", []):
+                deep_links.append((company.get("url"), company))
+                for paged_activity in paged_activity_data.values():
+                    deep_links += [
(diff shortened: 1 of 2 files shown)


redbox/redbox/api/format.py
```diff
@@ -85,6 +85,13 @@ def format_mcp_tool_response(tool_response, creator_type: ChunkCreatorType) -> t
             paged_results = [v for v in paged_data.values() if v is not None]
             for page in paged_results:
                 deep_links += [(item.get("url"), item) for item in page.get("result", {}).get("items", [])]
+        case "paged_composite":
+            for company, paged_activity_data in result.get("items", []):
```
lizzie-perry-dbt, line 89: Is it worth adding None guard?

lizzie-perry-dbt (approved):
Looks great, 1 minor comment, if you like only :)

redbox/redbox/api/format.py
hwixley: already one on line 66 :)

PR #1040, opened 2026-05-19
feat: move transcribe into websocket

nboyse:

- [ ] Yes

- [ ] No


django_app/frontend/src/js/web-components/chats/send-message-with-dictation.js
```diff
@@ -93,24 +93,41 @@ export class SendMessageWithDictation extends HTMLElement {
     this.isRecording = true;
     this.isStreaming = true;
     this.hideRecordButton();
+    this.lastFinalTranscript = "";
 
-    const client = new TranscribeStreamingClient({
```
hwixley, line 97: can we remove this import and package now ?

hwixley (commented):
nice!

should we deprecate the aws-transcribe package in this PR ? so can fix those vulns too

django_app/frontend/src/js/web-components/chats/send-message-with-dictation.js
nboyse: oh yes - good point. Have removed

david-okeke1337 (approved):

PR #1055, opened 2026-05-20
fix: remove write of metadata.document_schema for non-schematised indexes

hwixley:
## Context
The standard index's mappings are non-strict, causing unwanted mapping updates on metadata values. This resulted in an Opensearch BulkIndexError 'Limit of total fields [1000] has been exceeded' due to mapping automatically updating from unwanted metadata.

Particularly was ingesting `metadata.document_schema` meant only for schematised index.

## What
Adds loader flag to indicate whether schema should be ingested, preventing further mapping updates.

- [ ] Yes

- [ ] No


lizzie-perry-dbt (approved):
Looks good on fix - maybe one suggestion would be to adjust the tests in test_ingest ? Not necessary but if have time.

PR #1056, opened 2026-05-20
feat: add follow-up prompt recommendations for datahub agent

hwixley:
## Context
So users have recommendations for follow-up prompts on Datahub agent.

## What
Adds prompt to datahub_agent agent to provide follow-up prompt recommendations. Adds `mcp_tools` to prompt vars so model has access to available tools.

- [ ] Yes

- [ ] No

- Use datahub agent, verify provides follow-up prompts at end of response if not asking for user feedback


lizzie-perry-dbt (approved):
Looks good to me!
One optional comment would be maybe it is worth adding some tests (e.g. test that the DataHub prompt renders with mcp_tools, and test for when tools = []

PR #1060, opened 2026-05-21
feature/ASSIST-1502-conversation-structure

chopkinsmade:
## Context

PR to make changes for the Conversation structure section of the Assist Accessibility report

## What

- Make use of the aria-busy attribute to indicate to a screen reader that the chat with the LLM is ongoing. This value is set to `true` when the conversation starts and is set to `false` when either the response from the LLM ends or an error occurs
- Render the chats as an `li` inside an `ol` to help screen readers identify the order of the chats
- Add an aria-label to each chat to highlight to a screen reader who the message belongs to

- [ ] Yes

- [ ] No

### DIFF
--- a/django_app/frontend/src/js/chats.js
+++ b/django_app/frontend/src/js/chats.js
@@ -20,14 +20,14 @@ import "./web-components/documents/file-upload.js";
 import { syncUrlWithContent, refreshUI } from "./services";
 import { ChatHistory } from "./web-components/chats/chat-history.js";
 import { getActiveChatId } from "./utils/active-chat.js";
+import { Events, listenEvent } from "../interaction_design_system/ids/events/events.js";
 
-document.addEventListener("chat-response-start", (evt) => {
-  document.getElementById("chat-feed")?.setAttribute("aria-busy","true")
-})
+listenEvent(Events.CHAT_RESPONSE_START, (evt) => setAriaBusy(true));
+listenEvent(Events.CHAT_RESPONSE_ERROR, (evt) => setAriaBusy(false));
 
-document.addEventListener("chat-response-error", (evt) => {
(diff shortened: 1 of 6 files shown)


django_app/frontend/src/js/chats.js
```diff
@@ -21,6 +21,13 @@ import { syncUrlWithContent, refreshUI } from "./services";
 import { ChatHistory } from "./web-components/chats/chat-history.js";
 import { getActiveChatId } from "./utils/active-chat.js";
 
+document.addEventListener("chat-response-start", (evt) => {
+  document.getElementById("chat-feed")?.setAttribute("aria-busy","true")
+})
+
+document.addEventListener("chat-response-error", (evt) => {
+  document.getElementById("chat-feed")?.setAttribute("aria-busy","false")
+})
```
david-okeke1337: Can we use listenEvent for these, there should be examples in a few other js files. Makes it easier to change the EventTarget in the future and avoid any conflicts with other design systems (e.g. govuk-frontend)

django_app/redbox_app/templates/macros/chat-macros.html
```diff
@@ -28,15 +28,13 @@
...
 
 
-{% set role_text = message.role %}
+{% set role_text = "AI response" %}
 {% if message.role == "ai" %}
-{% set role_text = "Redbox" %}
-{% elif message.role == "user" %}
-{% set role_text = "You" %}
+{% set role_text = "User message" %}
 {% endif %}
```
david-okeke1337, line 36: This is working in reverse atm, the user message is getting labelled as "AI response" and the AI message is getting labelled as "User Message". We can probably tidy up this logic a bit as well.

david-okeke1337 (approved):

PR #1061, opened 2026-05-21
feat: optimise MCP session overhead by grouping mcp tool calls

hwixley:
## Context
MCP sessions add significant outbound HTTP overhead. Previously we have created new MCP sessions per MCP tool call, this is largely inefficient as we can leverage the same session for multiple tool calls.

This PR updates `ToolRunner` to group tool calls into different types; `sync`, `mcp_async`; and then dispatches grouped tool calls appropriately (sync calls are not grouped).

It also updates for better tool failure annotation in the response.

## What
1. Updates ToolRunner to group MCP tool calls.
2. Deprecates `wrap_async_tool` for new `execute_mcp_tools` which takes multiple tool calls as input
3. Better error handling and annotation across tool calls in run.

- [x] Yes

- [ ] No

Anything with tool calls should test normal sync tools, as well as, mcp async tools


redbox/redbox/graph/nodes/runner/runner.py
```diff
@@ -208,6 +402,25 @@ def parse(self, future: Future, metadata: dict) -> Optional[AIMessage]:
...
 
+        match request.result_type:
+            case tr_models.FutureResultType.MCP_ASYNC:
+                mcp_results, mcp_failures = response
+
+                results = {}
+                for tool_id, item in mcp_results.items():
+                    results[tool_id] = self.parse_response(
+                        future_tool_name=future_tool_name, response=item, is_intermediate_step=is_intermediate_step
+                    )
```
lizzie-perry-dbt: One bad MCP result aborts the whole group runner (calls parse_respons eis called which raises on any None/empty etc).
Maybe would be good to wrap the parse_response() in a try/except , so the good results still flows through

redbox/redbox/graph/nodes/runner/models.py
```diff
@@ -0,0 +1,108 @@
...
+
+    class Success(Base):
+        """Successful result of single tool execution."""
+
+        response: AIMessage = Field(description="AIMessage response generated from tool execution.")
+
+    class Failure(Base):
+        """Failed result of single tool execution."""
+
+        error: str = Field(default=None, description="Error from tool execution.")
```
lizzie-perry-dbt: Should default be Optional[str]

redbox/redbox/graph/nodes/runner/runner.py
```diff
@@ -166,32 +296,96 @@ def submit(self, tool_call: ToolCall) -> tuple[Future, dict]:
...
+    #         "intermediate_step": is_intermediate_step,
+    #         "tool_args": raw_args,
+    #         "future_result_type": tool_call.future_result_type,
+    #     }
+
+    def submit_mcp_async(
+        self, mcp_async_call: tr_models.ToolCallRequest.MCPAsync
+    ) -> tr_models.SubmittedToolCallRequest:
+        """Find, validate, and submit a grouped MCP async tool call to the executor. Returns (future, metadata)"""
+        is_intermediate_step = "False"
```
lizzie-perry-dbt: Is it always false ? It is hard coded here, should signal it is an intermediate step if it is, as in old code? But maybe it cannot be here?

redbox/redbox/graph/nodes/runner/wrap_async.py
```diff
@@ -23,105 +25,174 @@ def _get_mcp_headers(sso_access_token: str | None = None) -> dict[str, str]:
...
+            return asyncio.run(execute_mcp_tools_async(mcp_input))
         except Exception as e:
-            log.error(f"wrap_async_tool - Unhandled error running tool '{tool_name}': {e}", exc_info=True)
+            log.error(
+                f"execute_mcp_tools - Unhandled error when calling MCP server '{mcp_input.mcp_server}': {e}",
+                exc_info=True,
+            )
             raise
 
     return wrapper
```
lizzie-perry-dbt, line 202: Does this return the wrapper function? In description says should return a list of results?

redbox/tests/graph/nodes/runner/test_runner.py
```diff
@@ -697,35 +794,46 @@ def test_run_returns_correct_result(self, multi_tool_runner, mock_tool_b, tool_c
...
+                    result_type=tr_models.FutureResultType.SYNC,
+                    future=mock_future,
+                    future_args={},
+                    metadata={"intermediate_step": "False"},
+                )
+            ],
+            failures=[],
+        )
+        stub_result = tr_models.Result(
             responses=[
```
lizzie-perry-dbt: Should this be results rather than responses - I don't think Result() has a responses field?

lizzie-perry-dbt (approved):
Looks great!! It's a big PR so I could have missed some bits, just left a couple comments to take a look at.

Maybe worth adding test about the intermediate_step (in loop mode, is the intermediate_step= carried through?)

redbox/redbox/graph/nodes/runner/runner.py
hwixley: great shout

redbox/redbox/graph/nodes/runner/models.py
hwixley: good spot, left as non-optional but removed default

redbox/redbox/graph/nodes/runner/wrap_async.py
hwixley: bumped docstrings now

redbox/redbox/graph/nodes/runner/runner.py
hwixley: yes this is somewhat old code so was planning to refactor it soon, but yes have updated to actually set from tool calsl

PR #1066, opened 2026-05-27
fix: Planner produces plan contains 2 Evaluator tasks

lizzie-perry-dbt:
## Context

Planner produces plan contains 2 Evaluator tasks

In the current design, we expect the planner will produce the plan that contains no more than 1 Evaluator agent task,
e.g. 
Task1: agent: Evaluator
Task1: agent: Websearch, Task2: agent: Evaluator 
Task1: agent: Websearch, Task2: agent: Websearch, Task3: agent: Evaluator 

## What

Adapted prompt as well as the remove_evaluator_task func so that in the case that there are multiple evaluator tasks, all of them are removed and the task_evaluator is populated from the last evaluator task

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/redbox/graph/nodes/processes.py
+++ b/redbox/redbox/graph/nodes/processes.py
@@ -368,12 +368,12 @@ def remove_evaluator_task(state: RedboxState):
     if not state.agent_plans or not state.agent_plans.tasks:
         return state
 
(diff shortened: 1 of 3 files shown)


hwixley (approved):
Nice!

PR #1068, opened 2026-05-27
feat: Move BulkUserToolForm user search onto the server

david-okeke1337:

## Context

This PR moves the user search logic for the `BulkUserToolForm` onto the server and restricts the results size. This scales better as it prevents timeouts when querying from a large number of users (e.g. in prod). Also includes some small tweaks to the form error template (still needs design work) and adds a filter to the `ToolAccessRule` django-admin page.

## What

- Moved `BulkUserToolForm` user search logic to server.
- Updated form error template to properly output relevant information (field label/name and error description(s).
- Updated/Added tests.

- [x] Yes

- [ ] No

Note: local testing only covers basic functionality as it cannot mimic the number of users and data relationships in prod. Further performance testing can be carried out once merged/deployed to dev/prod.

- Ensure user search still works as expected
- Ensure search only happens once minimum 2 characters are entered
- Ensure empty/no-match messages show as and when expected

## Relevant links
image
image
image
image
image


django_app/redbox_app/redbox_core/views/tools_views.py
```diff
@@ -171,24 +171,33 @@ def dispatch(self, request, *args, **kwargs):
...
-                    "access_type": access_type,
-                },
-            )
+        existing_user_tools = UserTool.objects.filter(tool=self.tool, user_id__in=raw_user_ids)
+        existing_user_ids = set(existing_user_tools.values_list("user_id", flat=True))
+        existing_user_tools.update(role=role, access_type=access_type)
+
+        UserTool.objects.bulk_create(
+            [
+                UserTool(
```
chopkinsmade, line 190: You can avoid the extra call to the User table on line 180, and just pass in the id here as `user_id`. Django will work out your passing the id of an object and link it
```
UserTool(
                    user_id=user_id,
                    tool=self.tool,
                    role=role,
                    access_type=access_type,
                )
                for user_id in raw_user_ids
                if user_id not in existing_user_ids
```

django_app/redbox_app/redbox_core/views/tools_views.py
```diff
@@ -201,13 +210,36 @@ def get_context_data(self, **kwargs):
...
             }
         )
 
         return context
 
 
+class ToolUserSearchView(LoginRequiredMixin, ToolManagerRequiredMixin, View):
+    template_name = "tools/users/user-search-rows.html"
+
+    def get(self, request, slug):
```
chopkinsmade, line 221: This endpoint is getting hit from JS even when a single character is entered. It's good to have this limitation in endpoint, but it would be better to not send the request to the server in the first place

Image

django_app/redbox_app/redbox_core/forms.py
```diff
@@ -376,23 +376,15 @@ class Meta:
...
-
-        self.fields["user_ids"].queryset = unassigned_users
-
-    def clean_user_ids(self):
-        user_ids = self.cleaned_data["user_ids"]
+    def clean(self):
+        cleaned_data = super().clean()
 
-        users = self.unassigned_users.filter(pk__in=user_ids)
+        raw_user_ids = self.data.getlist("user_ids")
```
chopkinsmade, line 386: Can we leave the check in for ensuring a user selects at least 1 user? If I navigate to the bulk add page, and hit the "Add" button without making any selection I'm redirected back to the settings page

django_app/redbox_app/redbox_core/views/tools_views.py
david-okeke1337: good spot!

chopkinsmade (approved):

chopkinsmade (approved):

PR #1069, opened 2026-05-27
fix: Update search doc/knowledgebase related tools to not returned empty list.

lizzie-perry-dbt:
When performing query to Opensearch and no chunks were found, Assist will return an empty list which will cause the tool to fail causing the application to fail.

## What
- Fix the tools (both search for document, and knowledge base) to return a text indicating that no relevant chunks/information are found.
- When a user does not select any document i.e. s3_keys = [] , Assist should not call Opensearch but let the user know that they have not selected any document. Ideally, the planner should have identified that document is not selected and therefore search-related agents should not be used.
- Write relevant tests

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/redbox/graph/nodes/tools.py
+++ b/redbox/redbox/graph/nodes/tools.py
@@ -206,12 +206,17 @@ def build_search_documents_tool(
     no_selection_msg = (
         "No documents have been selected, so no search was performed. Ask the user to select at least one document."
     )
-
     no_results_msg = "No relevant information was found in the selected documents. Try to rephrase your prompt or proceed without information from the selected document(s)"
+    no_permission_msg = (
+        "User does not have permission to access documents. Ask the user to select a document they are permitted to use"
+    )
 
     def search_repo(query, selected_files, permitted_files, ai_settings, start_time=time.time()):
         if not selected_files:
             return no_selection_msg, []
+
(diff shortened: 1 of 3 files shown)


redbox/redbox/graph/nodes/tools.py
```diff
@@ -203,7 +203,16 @@ def build_search_documents_tool(
...
+    no_selection_msg = (
+        "No documents have been selected, so no search was performed. Ask the user to select at least one document."
+    )
+
+    no_results_msg = "No relevant information was found in the selected documents. Try to rephrase your prompt or proceed without information from the selected document(s)"
+
     def search_repo(query, selected_files, permitted_files, ai_settings, start_time=time.time()):
+        if not selected_files:
+            return no_selection_msg, []
+
```
hwixley, line 217: should we have an extra check on permitted files ?

ie. check if we have permissions and if not then return early with a specific unauthorised message

hwixley (approved):
woo! nice stuff on quick fix

small comment on potentially an extra permissions check

redbox/redbox/graph/nodes/tools.py
lizzie-perry-dbt: Absolutely! Have added this check now and comment to go with :)

PR #1074, opened 2026-05-29
feature/ASSIST-1507-name-role-value

chopkinsmade:
## Context

PR to make changes for the Name, role and value section of the Assist Accessibility report

## What

- Add an `aria-pressed` attribute to the feedback thumbs up and down buttons to highlight to a screen reader these elements are clickable, and their current clicked state
- Remove the `aria-expanded` attribute from the show-more elements that are not eligible to use it 
- Remove the `aria-readonly` attribute from the file upload `div` element as it is not eligible to use it 
- [ ] Yes

- [ ] No


david-okeke1337 (approved):
lgtm!

PR #1082, opened 2026-06-02
Search docs tool returned empty or whitespace-only response

lizzie-perry-dbt:
## Context

The search documents tool has suddenlt started failing due to the search documents tool returning an empty or whitespace only string.

failed: _search_documents - Tool '_search_documents' returned empty or whitespace-only response

## What

replicate error in unit tests

fix bug so that root cause is fixed

- [ ] Yes

- [ ] No


hwixley (approved):
nice stuff, very useful guardrails

think we can do a separate PR for other fixes for the blank error

PR #1094, opened 2026-06-04
feature/ASSIST-1511-dialog

chopkinsmade:
## Context

PR to make changes for the expandable menu section of the Assist Accessibility report

## What

On smaller screens, when the menu is expanded the elements behind the menu are still navigable using the tab button. This change adds some attributes to the menu to highlight to a screen reader the menu is acting as a modal, and sets the `inert` attribute on the div behind to block the elements being selected

- [ ] Yes

- [ ] No


david-okeke1337 (approved):
lgtm

PR #1102, opened 2026-06-05
feat: prepopulate chat message input from ?q= url param

joshwong-cs:
## Context

* data hub would like to link to dbt assist with a pre-filled prompt as a shortcut to allow users to access summaries of its companies/interactions
* thus enabling other services to do similar type of linking

## What

* adds support for pre-filling the chat message input via a `?q=` url query parameter on the chats page
* opens a new chat with `q`'s value already typed in the message input
* takes effect only on the new-chat route
* existing chats ignore the param so we don't clobber a user's in-progress draft
* output is auto-escaped so untrusted query params can't inject html

- [x] Yes

- [x] No


david-okeke1337 (approved):
lgtm!

joshwong-cs:
@david-okeke1337 leaving this for assist to merge, the PR still seems to be gated on the CI

joshwong-cs:
@david-okeke1337 @hwixley checking up on this, any idea why the dd gates aren't running (which i assume is the only blocker here?)

PR #1104, opened 2026-06-05
fix: LLM Error - Blank Response error, in stream_plan

lizzie-perry-dbt:
## Context

Error appearing in the logs which shows "LLM Error - Blank Response" - but we found no error in DataDog Traces.
With an empty messages list, the final_state.messages in app.py catches the exception and logs a the error, despite the stream_plan working

## What

We believe that the root of the error is that stream_plan showed the plan to the user but never saved it to state_messages, so the final state had no message, and therefore gave the blank response error.

stream_plan now returns the plan as an message, so therefore the final state has a message, stopping the blank message going through, and therefore error log message.

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/redbox/graph/nodes/processes.py
+++ b/redbox/redbox/graph/nodes/processes.py
@@ -881,7 +881,8 @@ def stream_plan():
         plan_parts.append(f"\n\n{suffix}")
         dispatch_custom_event(RedboxEventType.response_tokens, data=plan_parts[-1])
 
-        return {"messages": [AIMessage(content="".join(plan_parts))]}
+        state.messages = state.messages + [AIMessage(content="".join(plan_parts))]
+        return state
 
     return _stream_plan
 
--- a/redbox/redbox/graph/nodes/processes.py
+++ b/redbox/redbox/graph/nodes/processes.py
@@ -44,9 +44,9 @@ from redbox.models.chain import (
 )
 from redbox.models.graph import ROUTE_NAME_TAG, RedboxActivityEvent, RedboxEventType
 from redbox.models.prompts import (
-    USER_FEEDBACK_EVAL_PROMPT,
-    DATAHUB_USER_FEEDBACK,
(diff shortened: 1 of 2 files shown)


redbox/redbox/graph/nodes/processes.py
```diff
@@ -871,11 +871,17 @@ def stream_plan():
...
         for i, t in enumerate(state.agent_plans.tasks):
-            dispatch_custom_event(RedboxEventType.response_tokens, data=f"{i + 1}. {t.task}\n\n")
-        dispatch_custom_event(RedboxEventType.response_tokens, data=f"\n\n{random.choice(suffix_texts)}")
-        return state
+            plan_parts.append(f"{i + 1}. {t.task}\n\n")
+            dispatch_custom_event(RedboxEventType.response_tokens, data=plan_parts[-1])
+        plan_parts.append(f"\n\n{suffix}")
+        dispatch_custom_event(RedboxEventType.response_tokens, data=plan_parts[-1])
+
+        return {"messages": [AIMessage(content="".join(plan_parts))]}
```
hwixley: testing this on my local seems to cause issues in response when plan has more than one task

ie. "get weather in london and tokyo" - seems to only ever get results from one not both

seemed to fix it by changing `return state`, may be worth checking if can rather just return state

hwixley (changes requested):
nice stuff! think i just found an issue in some of the handling of plans with multiple tasks

redbox/redbox/graph/nodes/processes.py
lizzie-perry-dbt: Good spot Harry! I will check if that works ok now :)

lizzie-perry-dbt: Added the messages to the state and then returned state, and the tests to match the new return

hwixley (approved):
Nice ! working for me :)

PR #1105, opened 2026-06-05
feat: mcp tool cache and new run_tools_parallel_extended

hwixley:
## Context
Sets up MCP tool caching so we don't need to list tools for every request. This tool cache uses wrapped token as the cache key so that tool cache embed any token-based access control on MCP server.

Given adding caching means there is a case where we call the MCP server when it is down, have added `run_tools_parallel_extended` so that the datahub agent is aware of tool errors and returns appropriate error to user.

## What
- Adds token-based datahub MCP tool cache, cache lasts 5 minutes
- Adds `run_tools_parallel_extended` for agents to be slowly migrated to, to leverage better error handling

- [x] Yes

- [ ] No

1. Starting with MCP server on, make request to Datahub_Agent
2. Turn off MCP server
3. Run new request to Datahub_Agent
4. Check it returns response saying unable to contact MCP server

You need breakpoints to actually check whether cache is being used or not, or watch logs

### DIFF
--- a/redbox/redbox/api/wrapper.py
+++ b/redbox/redbox/api/wrapper.py
@@ -30,6 +30,14 @@ class SensitiveValue:
         memo[id(self)] = new
         return new
 
+    def __hash__(self) -> int:
+        return hash(self.get())  # hashed, never logged
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, SensitiveValue):
+            return NotImplemented
+        return self.get() == other.get()
+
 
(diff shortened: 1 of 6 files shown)


hwixley:
The PR was left as it took a backseat given the file ingestion refactor was prioritised, however, the implementation was working and mostly done. Think remaining thing is to review, test locally, and write/fix any outstanding tests.

lizzie-perry-dbt (dismissed):
Approved and made changes where failures were happening in test suits

PR #1106, opened 2026-06-08
feat: add admin opensearch file management

hwixley:
## Context
Adds action to file page in admin panel to check health of file ingestion. Allowing admins to select files and see if they were properly ingested in OpenSearch, if not or duplicates are found it has a button to reingest the file.

This is only supported for files which are already ingested in DB.

It is important to note health check page does not refresh results on page refresh, you have to redo "go" call in files page to reload results.

## What
- Fixes TextractChunkLoader to take chunk_resolution as input
- Fixes chunk strategy for largest chunks
- Adds action 'File Ingestion Health Check' to Files in admin panel
- Adds page 'File Ingestion Health Check' to view file ingestion health in admin panel, has a button to reingest

- [ ] Yes

- [ ] No

1. go to admin panel
2. go to "files"
3. select some files and action "File Ingestion Health Check"
4. click "go"
5. should be redirected to page with results from files
6. identify problematic files and click reingest
7. click back
8. click "go" again in files page
9. reingested file should be all green

### DIFF
--- a/redbox/redbox/admin/ingest.py
+++ b/redbox/redbox/admin/ingest.py
@@ -52,6 +52,17 @@ def _get_resolutions_for_file(file_uri: str, index_name: str) -> list[ChunkResol
     return resolutions.values()
 
 
+def _chunk_dedupe_key(hit: dict) -> tuple:
+    meta = hit["_source"]["metadata"]
+
+    return (
+        meta["uri"],
+        meta["chunk_resolution"],
+        meta["page_number"],
+        hit["_source"]["text"],
+    )
+
+
 def _get_duplicate_chunks(
     file_uri: str,
     index_name: str,
@@ -80,6 +91,7 @@ def _get_duplicate_chunks(
             },
             "_source": [
                 "text",
+                "metadata.uri",
                 "metadata.chunk_resolution",
                 "metadata.page_number",
             ],
@@ -92,24 +104,19 @@ def _get_duplicate_chunks(
     resolution_duplicate_counts: dict[str, int] = defaultdict(int)
 
     for hit in resp["hits"]["hits"]:
-        source = hit["_source"]
-        metadata = source["metadata"]
+        key = _chunk_dedupe_key(hit)
 
-        resolution = metadata["chunk_resolution"]
-        page_number = metadata["page_number"]
-        text = (source.get("text") or "").strip()
-
-        key = (
-            resolution,
-            page_number,
-            text,
-        )
-
-        if key in seen:
-            resolution_duplicate_counts[resolution] += 1
-            resolution_pages[resolution].add(page_number)
-        else:
+        if key not in seen:
             seen.add(key)
+            continue
+
+        meta = hit["_source"]["metadata"]
+
+        resolution = meta["chunk_resolution"]
+        page_number = meta["page_number"]
+
+        resolution_duplicate_counts[resolution] += 1
+        resolution_pages[resolution].add(page_number)
 
     return [
         ChunkDuplicateDetail(
--- a/redbox/redbox/admin/ingest.py
+++ b/redbox/redbox/admin/ingest.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from collections import defaultdict
 
 from redbox.models.settings import get_settings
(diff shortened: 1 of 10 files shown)


saisakul:
can you add/modify unit ingestion tests to check if correct chunks are ingested according to document types?

redbox/redbox/admin/ingest.py
```diff
@@ -0,0 +1,113 @@
...
+
+from redbox.models.settings import get_settings
+from redbox.models.file import ChunkResolution
+
+env = get_settings()
+
+
+@dataclass
+class ChunkResolutionDetail:
+    name: str
```
saisakul, line 12: can we add field description please?

redbox/redbox/admin/ingest.py
```diff
@@ -0,0 +1,113 @@
...
+    for res in expected_resolutions:
+        if res not in resolutions.keys():
+            resolutions[res] = ChunkResolutionDetail(name=res, count=0)
+
+    return resolutions.values()
+
+
+def _get_duplicate_chunks(file_uri: str, index_name: str) -> list[ChunkDuplicateDetail]:
+    """
+    Returns per-resolution duplicate summary
```
saisakul: can you please add more details doc sting to explain duplication definition/logic

redbox/redbox/loader/ingester.py
```diff
@@ -109,9 +112,10 @@ def _ingest_file(file_name: str, es_index_name: str = alias, enable_metadata_ext
     tabular_chunk_ingest_chain = ingest_from_loader(
         loader=TextractChunkLoader(
             bucket=env.bucket_name,
-            min_chunk_size=env.worker_ingest_min_chunk_size,
-            max_chunk_size=env.worker_ingest_max_chunk_size,
-            overlap_chars=0,
+            chunk_resolution=ChunkResolution.tabular,
```
scipio1465, line 116: There is a reason why tabular and larger are exactly the same chain? Then at L146 they both run

```
new_ids = RunnableParallel(
            {
                "normal": chunk_ingest_chain,
                "largest": large_chunk_ingest_chain,
                "tabular": tabular_chunk_ingest_chain,
                "schematised_tabular": tabular_schema_chunk_ingest_chain,
```
Also not sure why you want to do "normal" for csv/etc.. making embedding of tables. And probably you will never do semantic similarity with tabular data (I thiink!)

redbox/redbox/loader/ingester.py
```diff
@@ -83,6 +84,7 @@ def _ingest_file(file_name: str, es_index_name: str = alias, enable_metadata_ext
     chunk_ingest_chain = ingest_from_loader(
         loader=TextractChunkLoader(
             bucket=env.bucket_name,
+            chunk_resolution=ChunkResolution.normal,
             min_chunk_size=env.worker_ingest_min_chunk_size,
             max_chunk_size=env.worker_ingest_max_chunk_size,
             overlap_chars=0,
```
scipio1465, line 91: I'd suggest to have a env variable for the overlap, something like `worker_ingest_normal_chunk_overlap`

Not sure if some optimization has been done to set this to 0, but if in env variable this can be studied in the future.

redbox/redbox/loader/ingester.py
```diff
@@ -161,3 +166,36 @@ def ingest_file(file_name: str, es_index_name: str = alias) -> str | None:
...
+            "sort": [{"metadata.chunk_resolution.keyword": "asc"}, {"metadata.page_number": "asc"}],
+        },
+    )
+
+    seen: set[tuple] = set()
+    ids_to_delete: list[str] = []
+
+    for hit in resp["hits"]["hits"]:
+        meta = hit["_source"]["metadata"]
+        key = (meta["uri"], meta["chunk_resolution"], meta["page_number"])
```
lizzie-perry-dbt: Could it be possible that multiple chunks are across the same page, so would share a page number, so is the potential of deleting things we don't want to.
We could add index to the key so makes sure only the chunks we want to remove are removed

redbox/redbox/admin/ingest.py
```diff
@@ -0,0 +1,113 @@
...
+    # resolution -> list of per-page doc_counts
+    resolution_page_counts: dict[str, list[int]] = defaultdict(list)
+    after = None
+
+    while True:
+        composite = {
+            "sources": [
+                {"uri": {"terms": {"field": "metadata.uri.keyword"}}},
+                {"resolution": {"terms": {"field": "metadata.chunk_resolution.keyword"}}},
+                {"page_number": {"terms": {"field": "metadata.page_number"}}},
```
lizzie-perry-dbt: This could be same issue as other comment to do with the same page number and not using index, so could result in removing more than just a duplicate if something else is on the same page

redbox/redbox/loader/ingester.py
```diff
@@ -161,3 +166,36 @@ def ingest_file(file_name: str, es_index_name: str = alias) -> str | None:
...
+
+
+def remove_duplicate_chunks(file_uri: str, index_name: str) -> None:
+    es = env.elasticsearch_client()
+    log.info("Using Elasticsearch client: %s", es)
+
+    resp = es.search(
+        index=index_name,
+        body={
+            "size": 1000,
```
lizzie-perry-dbt, line 179: Does this onlt look at the first 1000 chunks - could there be more than this? - like in _get_duplicate_chunks which goes through everything

lizzie-perry-dbt (changes requested):
Great work getting through all of this. Looks really good, is a big PR so just put some comments on a couple bits that I saw, mostly to do with duplication and deleting duplications, just to make sure it is only deleting what it should and no more

hwixley:
This has been implemented by https://github.com/uktrade/redbox/pull/1147

closing due to already being done

PR #1147, opened 2026-06-23
fix: fix file and chunk ingestion with unstructured

hwixley:
## Context
Fixes issues with file ingestion and chunking.

## What
- Updates ingestion to be; extraction -> [chunking chains] (ie. normal, largest, tabular etc.); so that extraction is not duplicated across chains
- Updates all primary file extraction strategies (unstructured and textract) to return Unstructured elements so it can be used by downstream `chunk_by_title` for layout-aware chunking. If primary file extraction methods fail (ie. falls back to pymupdf) then will do layout-naive chunking.
- Updates PDFs to do 2 extractions: 1st for normal chunks which uses textract document analysis with fallbacks to unstructured (auto then fast) and pymupdf, and 2nd for largest chunks which just uses pymupdf
- Updates largest, tabular and schematised_tabular chains to use correct chunk size and overlap
- Adds new services for extraction and chunking for routing to downstream custom extractors/chunkers

### PDF Extraction Strategies
-  `textract_document_analysis` - textract document analysis. default for PDF extraction. 
- `textract_document_analysis_large` - chunked textract document analysis. default for handling large PDF extraction (> 5MB). It chunks the document into 200-page chunks, sends separate textract calls for each and then stitches them back together ahead of downstream chunking.
- `unstructured_auto` - unstructured partition with strategy 'auto'. the 1st fallback on PDF extraction if textract fails.
- `unstructured_fast` - unstructured partition with strategy 'fast'. the 2nd fallback on PDF extraction if unstructured_auto fails.
- `pymupdf` - fast, simple, and naive PDF extraction method. default if other strategies fail. If used implies downstream chunking is not layout-aware.

### Services
- `DocumentExtractionService` - routes to downstream extraction services, ie. unstructured, textract etc.
- `DocumentChunkingService` - routes to downstream chunking services, ie. unstructured chunk_by_title, page-by-page, joined pages, tabular etc.

### End-to-End Flow
`_ingest_file()` -> `DocumentExtractionService()` (extracts page contents) -> `MetadataExtraction()` (extracts metadata for page contents) -> `ingest_chunks(...)` (ingests chunks into opensearch with specified chunk config, using page contents)

#### Normal File (<=5MB)
1. File is ingested
2. Extract contents (if PDF uses fallback strategies for normal)
   - normal: strategy=textract_document_analysis (timeout 600s) -> strategy=unstructured_auto (timeout 240s) -> strategy=unstructured_fast (timeout 240s) -> pymupdf
   - largest: pymupdf
3. Chunk contents (appropriately chunks based on extraction type)
   - normal: if used textract or unstructured uses chunk_by_title, otherwise if pymupdf then page-by-page
   - largest: uses non-page-by-page chunking

#### Large File (>5MB)
1. File is ingested
2. Extract contents (if PDF uses large file textract strategy)
   - normal: strategy=textract_document_analysis_large (timeout 900s) -> pymupdf
   - largest: pymupdf
3. ... (same)

### Tabular ingestion flow
CSV and Excel files follow a dedicated ingestion path that preserves their tabular structure throughout indexing and retrieval.

During extraction, each worksheet (or the CSV itself) is parsed into a single logical table. Column names and data types are inferred to produce a document_schema, while the table contents are serialised into a text representation beginning with a table identifier (for example, <table_name>sheet1</table_name>) followed by the header row and data rows.

The extracted table is then passed to the tabular chunking service. Tables that fit within the embedding model limits are indexed as a single chunk. Larger tables are split only on row boundaries so that rows are never broken across chunks. Each chunk repeats the table header to preserve context for embeddings and is assigned a sequential metadata.index indicating its position within the original table.

Two tabular indices are populated:

Normal index – stores the table text and standard document metadata for semantic retrieval.

Schematised tabular index – stores the same table text together with the inferred document_schema. This separate index exists solely because the document_schema requires additional OpenSearch mappings that are not present in the standard tabular index.

Both indices contain identical table content and use the same chunking strategy; the only difference is the presence of the document_schema field in the schematised index.

At retrieval time, if multiple chunks belonging to the same table are returned, they are automatically reconstructed into a single logical table by grouping on the source document and table name, ordering by metadata.index, removing the repeated headers from subsequent chunks, and concatenating the remaining rows. This reconstruction is transparent to downstream consumers, which receive the original table as though it had never been split.

- [x] Yes

- [ ] No


saisakul:
please update `.env.example` to include new env vars

redbox/redbox/loader/chunking/chunkers/page_by_page.py
```diff
@@ -0,0 +1,47 @@
...
+from langchain_core.documents import Document
+
+from redbox.models.chain import GeneratedMetadata
+from redbox.loader.chunking.base import BaseDocumentChunker
+
+logger = logging.getLogger(__name__)
+
+
+class PageByPageDocumentChunker(BaseDocumentChunker):
+    def _chunk(self, pages: list[str]):
```
saisakul, line 18: please add some docstrings for each of the functions

redbox/redbox/loader/extraction/metadata.py
```diff
@@ -0,0 +1,108 @@
...
+                metadata.name = original_metadata.get("filename") or file_name
+
+            return metadata
+        except ValidationError as e:
+            logger.warning(e.errors())
+            return GeneratedMetadata(name=original_metadata.get("filename") or file_name)
+
+    def get_first_10k_chars(self, elements: list[Element] | list[str] | list[dict]) -> str:
+        if all(isinstance(p, str) for p in elements):
+            return "".join(p for p in elements)[:10_000]
```
saisakul: maybe should change this check if it is more than >10k then stop joining

redbox/redbox/loader/ingester.py
```diff
@@ -149,15 +185,23 @@ def _ingest_file(file_name: str, es_index_name: str = alias, enable_metadata_ext
...
-def ingest_file(file_name: str, es_index_name: str = alias) -> str | None:
+def ingest_file(file_name: str, es_index_name: str = alias) -> tuple[FileIngestionResponse | None, str | None]:
     try:
-        _ingest_file(file_name, es_index_name)
+        response = _ingest_file(file_name, es_index_name)
+        return response, None
     except Exception:
         logging.exception("Error while processing file [%s]", file_name)
-        return traceback.format_exc()
+        return None, traceback.format_exc()
```
lizzie-perry-dbt, line 211: This can permenantly mark files as error, as the excpetion catches IngestionAlreadyInProgressed, and returns None here, which is then passed into worker.py, which sets file.status = FileStatus.errored.
so when Django-Q retries while a file is already ingesting, that file will permenantly be marked as error. This is good for duplicates but could cause potential issue with the route of propogation

lizzie-perry-dbt (commented):
Looks good when been trying on the test env!! Just left one comment here which could leave files marked as error, as a side effect of some deduplication

redbox/redbox/loader/ingester.py
hwixley: updated now to re-raise

lizzie-perry-dbt (approved):

PR #1148, opened 2026-06-25
fix: Fallback for Throttling Not working

lizzie-perry-dbt:
## Context

When there is a throttling issue i.e. `botocore.errorfactory.ServiceUnavailableException: An error occurred (ServiceUnavailableException) when calling the InvokeModelWithResponseStream operation (reached max retries: 4): Too many connections, please wait before trying again.` Assist should be using the fallback model. However, the current logic is not working and the fallback model is not called hence causing app to fail.

## What
 Root cause of the issue - the fallback logic in get_chat_llm wrapped init_chat_model in a try/except loop, but the get_chat_llm creates a python wrapper object, and does not many any API calls. So throttling errors surface later on at invocation, where the try/except has already exited and the exception is not caught.

The change here puts the fallback at the model level with langchain with_fallbacks, so fallback errors are caught when they happen.

- [ ] Yes

- [ ] No

### DIFF
--- a/redbox/redbox/chains/components.py
+++ b/redbox/redbox/chains/components.py
@@ -43,7 +43,7 @@ _FALLBACK_CACHE = {}
 FALLBACK_COOLDOWN_SECS = 420  # 7 mins feels ok
 
 _THROTTLING_CODES = frozenset(
-    {"ServiceUnavailableException", "Throttling_Exeption", "RateLimitExceed", "TooManyRequestsException"}
+    {"ServiceUnavailableException", "Throttling_Exeption", "RateLimitExceeded", "TooManyRequestsException"}
 )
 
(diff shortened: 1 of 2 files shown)


redbox/redbox/chains/components.py
```diff
@@ -40,6 +42,41 @@
 
 FALLBACK_COOLDOWN_SECS = 420  # 7 mins feels ok
 
+_THROTTLING_CODES = frozenset(
+    {"ServiceUnavailableException", "Throttling_Exeption", "RateLimitExceeded", "TooManyRequestsException"}
```
hwixley: NIT: typo on `Throttling_Exception` (missing the 'c')

hwixley (approved):

PR #1150, opened 2026-06-25
feat: add local db setup cmd

JasperWagen:
## Context

create an easy way to automatically set up required objects in the db

- [x] Yes

Delete these objects from your db and use the cmd to recreate

## Relevant links


django_app/redbox_app/redbox_core/management/commands/local_db_init.py
```diff
@@ -0,0 +1,32 @@
...
+        if not Environment.is_local:
+            raise NotLocalError
+
+        self.stdout.write("Making migrations...")
+        call_command("makemigrations")
+
+        self.stdout.write("Running migrations...")
+        call_command("migrate")
+
+        self.stdout.write("Loading tools...")
```
chopkinsmade, line 20: Happy to discuss this further, but if we moved the `loaddata` call from 3 separate commands into a single command and a single file we'd get transaction support from django? If the 2nd call to loaddata fails do we end up in a stuck state?

django_app/redbox_app/redbox_core/management/commands/local_db_init.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Command(BaseCommand):
+    help = "Initialise db with data from fixtures"
+
+    def handle(self, *_args, **_options):
+        if not Environment.is_local:
+            raise NotLocalError
+
+        self.stdout.write("Making migrations...")
+        call_command("makemigrations")
```
lizzie-perry-dbt: I don't believe "makemigrations" is a db setup step, so may need to omit this as running it could generate migration files in working tree

django_app/redbox_app/redbox_core/management/commands/local_db_init.py
```diff
@@ -0,0 +1,32 @@
...
+        self.stdout.write("Running migrations...")
+        call_command("migrate")
+
+        self.stdout.write("Loading tools...")
+        call_command("loaddata", "tools")
+
+        self.stdout.write("Loading tool settings...")
+        call_command("loaddata", "tool_settings")
+
+        self.stdout.write("Loading tool settings...")
```
lizzie-perry-dbt: Log command here should be "Loading agents"

django_app/tests/management/test_local_db_init.py
```diff
@@ -0,0 +1,51 @@
...
+def mock_local_env():
+    with patch.object(Environment, "is_local", new=True):
+        yield
+
+
+@pytest.fixture(autouse=True)
+def mock_migrations():
+    "Don't run the migrations at the start of the command to reduce test time"
+
+    with patch("django.core.management.call_command") as mock:
```
lizzie-perry-dbt: Patching "django.core.management.call_command" changes the function in Django namespace, but as local_db_init.py has it's own copy in there and does not check django namespace again, it will call the real function instead of the mock.
If patch to the below then patch to the copy that is in local_db_init.py

```suggestion
    with patch("redbox_app.redbox_core.management.commands.local_db_init.call_command") as mock:
```

lizzie-perry-dbt (changes requested):
Hiya - looks good here!! Nice work :) 
I have left a couple of small comments if you would like to take a look. Happy for you to push back if do not agree with any, of course!

PR #1168, opened 2026-07-02
feature/ASSIST-1566-run-app-without-external-dependancies

chopkinsmade:
## Context

Refactor the code to allow the integration tests to run using playwright
## What

Run the integration tests on a PR being raised. This PR has commented out the majority of the integration tests, and focussed on a simple test where the app is loaded, a user is signed in using the mock sso, and they are sent to the chats page. Future PRs will work through the `django_app/tests/playwright/test_journey.py` file re-adding additional scenarios

The makefile command that runs the playwright tests in `test-integration` is configured to always produce a video and screenshots to help debugging. There is a separate makefile command `test-integration-debug` for local that runs playwright in a headed version that can be used for local step by step debugging

- [ ] Yes

- [ ] No


django_app/redbox_app/settings.py
```diff
@@ -184,6 +184,7 @@
 ACCOUNT_EMAIL_VERIFICATION = "none"
 
 AUTHBROKER_URL = env.str("AUTHBROKER_URL", "https://example.com")
+# AUTHBROKER_INTERNAL_URL = env.str("AUTHBROKER_INTERNAL_URL")
```
JasperWagen: comment

django_app/tests/playwright/pages.py
```diff
@@ -46,7 +46,8 @@ def __init__(self, page: Page):
         self.page = page
         self.axe = Axe()
         self.check_title()
-        self.check_a11y()
+        # fix this test as it is waiting on elements not always present (the hideFooter variable)
```
JasperWagen: comment

django_app/tests/playwright/pages.py
JasperWagen: comment

tests/.env.integration
```diff
@@ -1,57 +1,67 @@
-EMBEDDING_BACKEND=amazon.titan-embed-text-v2:0
+# === Development ===
+DEV_MODE=true
```
JasperWagen, line 2: I feel like these should be stored in github actions rather than in a file. Did you have a reason for doing it this way?

chopkinsmade: A separate ticket will be created to tidy up the multiple .env files, as there is too much duplication

JasperWagen (approved):

PR #1171, opened 2026-07-06
fix: decrease token budget for tabular chunks

hwixley:
## Context
To prevent embedding overflow error on tabular chunks. Token budget is 8192 but still overflowed indicating a deviation in the tokeniser. This serves as a simple fix.

## What
Decreases token quota for tabualr chunks to prevent overflow.

- [ ] Yes

- [ ] No


redbox/redbox/loader/chunking/chunkers/tabular.py
```diff
@@ -20,7 +20,7 @@ class TabularDocumentChunker(BaseChunker):
     """
 
     MAX_CHARS = 50_000
-    MAX_TOKENS = 8_000
+    MAX_TOKENS = 6_000
```
saisakul: can we put the settings in `Settings` class please

saisakul (approved):

PR #1176, opened 2026-07-07
fix: add unstructured tesseract to path

hwixley:
## Context
Adds unstructured tesseract binary to docker to fix files that fallback to unstructured partition (ie. anything that is not normal file suffix ie. not pdf, docx, pptx, xlsx, csv, md, or html)

## What
Adds binary.

This has not been tested yet due to limitations on getting a docker build on local. I suggest testing this on test env before merging.

- [ ] Yes

- [ ] No

### DIFF
--- a/django_app/Dockerfile
+++ b/django_app/Dockerfile
@@ -51,9 +51,10 @@ RUN apt-get update && apt-get install -y \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


hwixley:
Has not been tested in docker build yet

Would suggest someone review and test the PR, to ensure file extraction works for unstructured strategy, and make any further changes if needed

lizzie-perry-dbt (approved):

PR #1177, opened 2026-07-07
fix: setup real bedrock tokeniser

hwixley:
- [ ] Yes

- [ ] No


hwixley:
this would add significant cost

rather best to implement a pessimistic local tokeniser that doesnt call aws

PR #1178, opened 2026-07-07
fix: deprecate pymupdf extraction for largest chunks

hwixley:
## Context
Deprecates default usage of pymupdf extraction for PDF largest chunks, rather use same textract OCR used in normal chunks.

This was due to an edge case on certain PDFs which cannot be parsed by pymupdf (ie. ones with no extractable text that need OCR).

## What
Deprecates pymupdf extraction usage for pdf largest chunks. Uses default texrtact for both largest and normal.

- [ ] Yes

- [ ] No


redbox/redbox/loader/ingester.py
```diff
@@ -101,15 +100,11 @@ def _ingest_file(
     # -- extraction --
     # normal chunk
     normal_extraction_strategy, normal_elements = extraction_service.extract(
-        file_name=file_name, chunk_resolution=ChunkResolution.normal
+        file_name=file_name,
     )
 
     # largest chunk
```
lizzie-perry-dbt: slightly out of date comment
```suggestion
    # largest chunk reuses normal extraction
```

redbox/redbox/loader/extraction/service.py
```diff
@@ -309,9 +307,6 @@ def _extract_pdf(
             file_size:
                 Size of the PDF in bytes.
```
lizzie-perry-dbt: Out of date docstring

pdf:

lizzie-perry-dbt (approved):
Looks good to me! Just some minor comments on out of date comments and doc string

PR #1188, opened 2026-07-10
feature/auto-create-minio-bucket

chopkinsmade:
## Context

Create a bucket with the `BUCKET_NAME` env var inside minio when container starts
## What

When the app starts, if you haven't already created the bucket in minio you will see bucket not found errors in the logs.

- [ ] Yes

- [ ] No


lizzie-perry-dbt (approved):
Nice work!

PR #1253, opened 2026-08-12
bugfix: strip out emails from citation source

nboyse:

- [ ] Yes

- [ ] No


saisakul (approved):

django_app/redbox_app/redbox_core/models.py
```diff
@@ -1544,7 +1544,7 @@ def internal_url(self) -> URL:
 
     @cached_property
     def display_name(self) -> str:
-        return str(self.uri) if not self.file else self.file.file_name
+        return str(self.uri).rsplit("/", 1)[-1] if not self.file else self.file.file_name
```
saisakul: would be good if we can check if it is an email before we stripped. as it could be a website

